Issue #171, opened 2026-07-09
P0: Define trusted time model for intent freshness, replay, velocity, and tool windows


AngeYobo:
## Summary

Define a trusted time model for OxDeAI authorization, replay protection, velocity limits, and tool-call windows.

A local policy-boundary repro showed that attacker-controlled `intent.timestamp` currently drives several critical controls:

- authorization `issued_at`
- authorization `expiry`
- replay nonce window eviction
- velocity window reset
- tool-call window reset

Because the agent controls the intent, using `intent.timestamp` as the effective clock lets an adversarial agent future-date intents and bypass time-window controls.

### Future-dated authorization

Input:

```text
intent.timestamp = now + 100 years
````

Observed:

```text
decision: ALLOW
issued_at: far future
expiry: far future + 60s
```

Impact:

An authorization can be minted with a far-future validity window if the verifier only checks `now < expiry`.

### Replay bypass by future dating

Control:

```text
nonce 7 first use        -> ALLOW
nonce 7 immediate replay -> DENY / REPLAY_NONCE
```

Attack:

```text
nonce 7 replay with timestamp beyond replay window -> ALLOW
```

Impact:

Replay protection can be bypassed because nonce eviction is based on attacker-controlled intent time.

### Velocity and tool-call window bypass

With `max_actions = 1` or `max_calls = 1`, sequential actions future-dated beyond the window all return:

```text
ALLOW
```

Impact:

Windowed rate limits can be reset by attacker-controlled timestamps.

## Root cause

The engine currently treats `intent.timestamp` as the effective time source for multiple security decisions.

For a protocol whose threat model includes adversarial or compromised agents, intent-supplied time must not be trusted without a bounded freshness model.

## Goal

Define the correct trusted-time architecture before implementing a patch.

This should specify how OxDeAI handles:

* intent freshness
* authorization issuance time
* authorization expiry
* replay nonce retention
* velocity windows
* tool-call windows
* deterministic evaluation
* state-provider responsibilities
* verifier responsibilities
* clock skew tolerance

## Required design questions

The issue should resolve:

1. What is the trusted time input?

Options may include:

* verifier-supplied `now`
* PEP-supplied `now`
* state-provider time
* policy-engine evaluation time
* deterministic external time parameter

2. How is clock skew handled?

Example:

```text
maxClockSkewSeconds = 300
```

Reject if:

```text
intent.timestamp > trusted_now + maxClockSkew
[...]

PR #174, opened 2026-07-11
Add draft trusted-time conformance vectors (#171)

Whatsonyourmind:
Adds a data-first draft conformance fixture for the trusted-time model discussed in #171, as scoped in the thread.

**What this adds**
- `packages/conformance/vectors/trusted-time.json` — six V/P vectors (`tt-neg-001..003`, `tt-pos-001..003`) matching the existing `packages/conformance/vectors/` schema/style.

**Semantics**
- `intent.timestamp` is a bounded freshness claim only; `evaluation_time` is the trusted clock used for issuance, expiry, replay retention, and velocity/tool-call windows.
- Negative/positive pairs pin both sides: V1/P1 freshness gate, V2/P2 replay retention, V3/P3 velocity windows. `tt-neg-003` vs `tt-pos-003` is the discriminating pair (same "window looks expired" surface, opposite trusted-time truth).
- Each vector is a pure `evaluate(intent, state, evaluation_time)` call; `input.preconditions` carry prior accepted events stamped at trusted time.

**Scope (intentionally narrow)**
- Adds one JSON file only. No implementation changes, no runtime behavior changes, no spec-wide rewrite.
- `version: 0.1.0-draft`; reason codes carry `"provisional": true` pending the reason-code enum lock.

Refs #171. Happy to adjust the shape/fields to match review.


AngeYobo:
Thanks !!  this is well scoped.

I’m away from my local machine at the moment, so I won’t merge this
immediately.

I’ll review it properly once I’m back, especially:

- whether the file is accepted by the current conformance validator
- whether the draft/provisional markers fit the existing vector schema
- whether it should stay under `packages/conformance/vectors/` or move
under a draft/non-normative path until #171 is locked

The scope looks correct:

- one JSON vector file
- no implementation changes
- no runtime behavior changes
- provisional reason codes
- trusted-time semantics only

The important part is preserving the semantics:

- `intent.timestamp` is a freshness claim only
- `evaluation_time` is the trusted clock
- replay/velocity behavior is keyed off `evaluation_time`
- V/P pairs prove both deny and allow paths

Thanks for keeping this narrow.

Whatsonyourmind:
Ran your review question ahead of time so it's one less thing when you're back at your machine — **does the current conformance validator accept this file?**

Short answer: it's **inert by construction**, which for a draft is the safe state. `validate.ts` discovers vectors by explicit filename — every check does `loadJson<VectorFile>("<name>.json")`, and the `loadJson` helper is just `resolve(process.cwd(), "vectors", name)` + `readFileSync` (no directory scan). Since nothing calls `loadJson("trusted-time.json")`, the file is never loaded: it can't fail CI, and it also isn't executed yet. So it stages cleanly without touching the existing suite — which matches leaving #171 open until the spec settles.

To *activate* it later is one addition: a `verifyTrustedTime` check that loads the file and maps each vector's `expected.decision / status / violations` onto a real `evaluate(intent, state, evaluation_time)` call, registered alongside the others. One thing to flag while you're there — it overlaps the existing `clock-semantics-verification.json` check, so the open design question is whether trusted-time *extends* that vector set or stands on its own. That's exactly the spec / implementation-path call you're keeping #171 open for, so no need to decide it here.

Happy to add the `verifyTrustedTime` wiring whenever it's useful — either folded into this PR if you'd rather land it active, or as a follow-up once the path is settled. Your call; no rush.

AngeYobo:
Thanks for checking that ahead of time. That’s helpful.

The fact that the draft vectors are inert by construction is actually the
behavior I was hoping for at this stage. I want #171 to remain a
design/spec discussion until the trusted-time model is finalized.

I also agree the question of whether this extends the existing clock
semantics or becomes its own verifier is a protocol design decision, so I’d
rather settle that first than wire it prematurely.

Once the trusted-time model is locked, I’d definitely welcome a follow-up
PR for the conformance wiring. Keeping the implementation separate from the
semantics is the right separation for now.

AngeYobo:
Thanks,  the PR scope and overall fixture structure are good.

I found one semantic issue we should resolve before merge: tt-neg-002 and tt-neg-003 both use intent timestamps beyond evaluation_time + maxClockSkewSeconds.

Under the proposed model, the freshness gate should reject those intents before replay or velocity evaluation. That means the expected reason in those two vectors would be INTENT_FRESHNESS_FUTURE, not REPLAY_NONCE or VELOCITY_EXCEEDED.

I think we should separate the concerns:

* future-dated replay/velocity attempts should prove that freshness rejection happens first;
* replay retention should be tested with a fresh timestamp and a reused nonce still retained under trusted evaluation_time;
* velocity should be tested with a fresh timestamp while trusted evaluation_time remains inside the active window.

That makes the evaluation order explicit and avoids expecting downstream reason codes for inputs that should already fail the freshness gate.

Everything else about the scope looks aligned. If you could adjust those two vectors accordingly, I think this draft fixture will be in good shape to merge.

Whatsonyourmind:
Good catch — you're right, and it's exactly the ordering the model implies: with freshness evaluated before replay/velocity, both of those intents (future-dated past `maxClockSkew`) would deny with `INTENT_FRESHNESS_FUTURE` and never reach the downstream gate, so asserting `REPLAY_NONCE` / `VELOCITY_EXCEEDED` on them was wrong.

Pushed the restructure (separating the concerns as you laid out):

- **tt-neg-002** → fresh timestamp (`= evaluation_time`) + nonce `n7` still retained → `REPLAY_NONCE`. Replay in isolation.
- **tt-neg-003** → fresh timestamp + trusted `evaluation_time` still inside the active window → `VELOCITY_EXCEEDED`. Velocity in isolation.
- **tt-neg-004** (new) → future-dated timestamp + the same retained nonce `n7` → `INTENT_FRESHNESS_FUTURE`. This is the ordering proof: an intent that would *also* be a replay still fails freshness first, so the gate order is explicit in the fixtures rather than implied.

The three positives were already fresh, so they're unchanged, and I added a sentence to the file description stating the evaluation order (freshness before replay/velocity). Semantics double-checked: the two isolated vectors are now within `maxClockSkew`, and only the freshness vectors (001, 004) are future-dated. Ready for another look.

AngeYobo:
Thanks, this is the right correction.

The new split is much clearer:

- `tt-neg-002` isolates replay semantics
- `tt-neg-003` isolates velocity semantics
- `tt-neg-004` pins the evaluation order by proving freshness rejection takes precedence over replay

Adding the evaluation-order statement to the file description is also the right move.

I’ll review the updated file and run the conformance checks once I’m back at my local machine. From the semantics you described, this now looks aligned.

AngeYobo:
This looks good to merge , thanks for the clean split on the evaluation order, the tt-neg-004 ordering proof and the tt-neg-003/tt-pos-003 pair are exactly right.

One tiny thing before I merge, purely for self-documentation: `maxIntentAgeSeconds` is declared in `config` but no vector exercises the stale (too-old) path yet. That's intentional, it's already planned for the follow-up pass (INTENT_STALE + a matching stale V/P pair)  but as it stands a reviewer coming to the file without this thread (e.g. during the OpenZeppelin pass) could read the unused config field as an oversight.

Could you add a one-line note to the file `description` so the draft reads correctly on its own? Something like:

> Stale-intent rejection (maxIntentAgeSeconds) is intentionally deferred to the follow-up vector pass; only the future-dating freshness direction is covered here.

That's the only change needed on my side. Happy to merge once it's in. Thanks again for keeping this so tightly scoped.

Whatsonyourmind:
Done — pushed in `4cae2eb`. Added the note to the file `description`, close to your wording so the draft stands on its own:

> Stale-intent rejection (maxIntentAgeSeconds) is intentionally deferred to the follow-up vector pass; only the future-dating freshness direction is covered here.

It sits right after the evaluation-order sentence, so a reviewer arriving cold (e.g. during the OpenZeppelin pass) reads the unused `maxIntentAgeSeconds` as a deliberate placeholder for the `INTENT_STALE` + stale V/P pair, not an oversight. Good call — that ambiguity would've cost someone a double-take.

Ready when you are. Thanks for the careful reviews on this one.

AngeYobo:
Thanks !!

AngeYobo:
Merged !! Thanks again

Whatsonyourmind:
Thanks @AngeYobo — glad it's in. Whenever you're ready (genuinely no rush), two natural next pieces are teed up and I'm happy to take either:

1. **Normative spec** — I left the inline draft up in #171 (the two-clock trust model, the `evaluate(intent, state, evaluation_time)` contract, the four consistency rules, and the reason-code set reconciled against the `ReasonCode` enum in `packages/core/src/types/policy.ts`). On your nod or edits I'll turn it into a narrow docs/spec PR — tightly scoped, DCO, matching your docs path.
2. **The deferred stale-intent pass** we flagged in the file description — `INTENT_STALE` + a matching stale V/P pair for the `maxIntentAgeSeconds` direction, mirroring the future-dating freshness case already covered, as a small follow-up vectors PR.

Whichever you'd rather see first, or both. Thanks for the quick, careful reviews on this one.

AngeYobo:
Thanks!  both are useful.

Let’s take the normative spec first.

Please keep it as a narrow `docs/spec/` PR covering only:

- the two-clock trust model
- the `evaluate(intent, state, evaluation_time)` contract
- the four consistency invariants
- required vs profile-defined configuration fields
- future and stale freshness semantics
- the proposed reason-code mapping against the existing `ReasonCode` enum,
with any additions or reuse remaining subject to review

No verifier wiring, runtime changes, or vector updates in that PR.

Once the spec text and reason-code mapping are agreed, the stale-intent V/P
pair can follow as a separate small vectors PR.

One reviewed step at a time is best here. Thanks for keeping the
contribution disciplined.

PR #175, opened 2026-07-13
docs(spec): add trusted-time authorization core profile v1

Whatsonyourmind:
Per the agreed scope in #171 — the normative trusted-time spec first, one reviewed step at a time.

Adds `docs/spec/core/trusted-time-v1.md` (indexed under Core in `docs/spec/README.md`). Scope is **spec only** — no verifier wiring, no runtime changes, no vector updates in this PR, exactly as discussed.

Covers, and only:

- **Two-clock trust model** — untrusted `intent.timestamp` (a bounded freshness claim only) vs trusted `evaluation_time` (sole authority for issuance, expiry, replay retention, and velocity).
- **`evaluate(intent, state, evaluation_time)` contract** — pure function of its inputs; `ALLOW` mints `issued_at`/`expiry`, `DENY` mints nothing and carries reason codes.
- **Four consistency invariants** — skew is freshness-only; trusted issuance; trusted + bounded expiry; downstream enforcement stays zero-tolerance.
- **Configuration** — REQUIRED (`maxClockSkewSeconds`, `maxIntentAgeSeconds`, `replayWindowSeconds`, `maxTtlSeconds`) vs profile-defined (`velocity.*`).
- **Freshness semantics** — future and stale, evaluated **before** replay/velocity.
- **Reason-code mapping** against the existing `ReasonCode` enum: **reuse** `REPLAY_NONCE` / `VELOCITY_EXCEEDED` / `AUTH_EXPIRED`; **propose** `INTENT_FRESHNESS_FUTURE` + `INTENT_STALE` — additions/reuse left **subject to your review** (including the two-codes-vs-parametrized question).

The reason-code mapping in §8 is the main thing to agree on. **No enum change is included** — that stays a separate reviewed step.

Once the spec text and the mapping are agreed, the stale-intent negative/positive vector pair follows as a separate small vectors PR (then `verifyTrustedTime`, then engine) — one reviewed step at a time.

### DIFF
--- a/docs/spec/README.md
+++ b/docs/spec/README.md
@@ -18,6 +18,7 @@ Execution is only reachable if authorization is verified.
 
 - canonicalization-v1
 - eta-core-v1
+- trusted-time-v1
 
 ### Artifacts
 
--- a/docs/spec/core/trusted-time-v1.md
+++ b/docs/spec/core/trusted-time-v1.md
@@ -114,8 +114,7 @@ A conformant trusted-time evaluation MUST hold all four invariants:
 2. **Trusted issuance.** `issued_at` MUST be minted from trusted time:
    `issued_at = evaluation_time`.
 3. **Trusted, bounded expiry.** `expiry` MUST be derived from trusted time and a
-   bounded TTL: `expiry = evaluation_time + min(maxTtlSeconds, requested_ttl)`.
-   Absent an optional `requested_ttl`, `expiry = evaluation_time + maxTtlSeconds`.
+   bounded TTL: `expiry = evaluation_time + maxTtlSeconds`.
 4. **Enforcement stays zero-tolerance.** Trusted-time *mints* the expiry;
    downstream expiry verification remains zero-tolerance (`now < expiry`),
    unchanged from `eta-core-v1`. This profile does not relax enforcement.
@@ -139,27 +138,7 @@ A conformant trusted-time evaluation MUST hold all four invariants:
   for a deployment that runs more than one PEP without a single authoritative
   clock (§2.1).
 
-### 5.1 `requested_ttl`
-
-`requested_ttl` appears in the expiry formula (§4.3) and is defined here:
-
-- **Source.** An OPTIONAL per-request desired TTL (carried on the request /
-  intent or supplied by policy). It is **not** a configuration field.
-- **Type / bounds.** A finite integer in `[1, maxTtlSeconds]`, seconds. `0` is
-  invalid — it would mint `expiry == issued_at`. The `min(maxTtlSeconds,
-  requested_ttl)` clamp (§4.3) is the enforced upper bound, and `maxTtlSeconds`
-  itself MUST be `> 0` (§5.2), so a valid clamp result is always `≥ 1`.
-- **Trust status.** `requested_ttl` MAY originate from an untrusted source. It is
-  trust-safe **only** because it enters through the `min()` clamp: it can shorten
-  the minted TTL but can never extend it beyond `maxTtlSeconds`. It MUST NOT be
-  used for any other time-dependent decision.
-- **Absent / invalid.** Absent → `expiry = evaluation_time + maxTtlSeconds`
-  (§4.3). Present but outside `[1, maxTtlSeconds]` (including `0`, negative,
-  non-integer, or non-finite; see §5.2) → the evaluation MUST fail closed and
-  `DENY` with `STATE_INVALID`; it MUST NOT be silently coerced to a permissive
-  value.
-
-### 5.2 Numeric Domain
+### 5.1 Numeric Domain
 
 All time and duration quantities MUST lie in a well-defined numeric domain;
 values outside it MUST cause the evaluation to fail closed, never a silent
@@ -169,9 +148,9 @@ coercion:
   within the safe-integer range (unix seconds).
 - Every duration field — `maxClockSkewSeconds`, `maxIntentAgeSeconds`,
   `replayWindowSeconds`, `maxTtlSeconds`, `maxInterPepSkewSeconds`,
-  `requested_ttl`, `velocity.windowSeconds` — MUST be a finite non-negative
+  `velocity.windowSeconds` — MUST be a finite non-negative
   integer; `velocity.maxActions` MUST be a finite non-negative integer count.
-  `maxTtlSeconds` and `requested_ttl` MUST additionally be `≥ 1` (a zero TTL
+  `maxTtlSeconds` MUST additionally be `≥ 1` (a zero TTL
   would mint `expiry == issued_at`).
 - A malformed `intent.timestamp` (NaN, Infinity, non-integer, negative, or
   unsafe magnitude) MUST `DENY` with `STATE_INVALID` at the freshness gate (§6),
@@ -184,6 +163,11 @@ coercion:
 
 Let `Δ = intent.timestamp − evaluation_time`. The freshness gate is evaluated
 **before** replay and velocity (§7 ordering), and is the only gate that reads
+`intent.timestamp`.
+
+**Precondition.** A malformed `intent.timestamp` (NaN, Infinity, non-integer,
+negative, or unsafe magnitude) fails closed with `STATE_INVALID` per §5.1
+*before* `Δ` is computed; the freshness comparisons below assume a well-formed
 `intent.timestamp`.
 
 - **Future-dated.** `Δ > maxClockSkewSeconds` → `DENY`, no authorization minted.
@@ -202,12 +186,10 @@ future-dated or stale — the freshness decision is reached first.
 Replay and velocity MUST key off `evaluation_time` only.
 
 - **Replay.** A nonce's retention/eviction MUST be computed against
-  `evaluation_time`, over the **effective retention window defined in §2.1** —
-  `[seen_at, seen_at + replayWindowSeconds)`, widened to
-  `[seen_at, seen_at + replayWindowSeconds + maxInterPepSkewSeconds)` under
-  bounded multi-PEP skew. §2.1 is the single normative definition of that window;
-  this section does not restate it. A future-dated `intent.timestamp` MUST NOT
-  evict a nonce still inside it. Reuse of a retained nonce → `DENY`.
+  `evaluation_time`, over the **effective retention window defined in §2.1**,
+  which is the single normative definition of that window; this section does not
+  restate the formula. A future-dated `intent.timestamp` MUST NOT evict a nonce
+  still inside it. Reuse of a retained nonce → `DENY`.
 - **Velocity.** The action window is
   `[window_start, window_start + velocity.windowSeconds)` in `evaluation_time`.
   A future-dated intent MUST NOT reset or advance it. Exceeding `maxActions`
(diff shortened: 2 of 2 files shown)


AngeYobo:
Thanks, the structure is strong and the lifecycle separation is clear.

Before merge, I think the profile needs a few additions to fully capture the model already discussed in #171:

1. Please specify the PEP time-source requirements:
   - evaluation_time captured once per evaluation
   - monotonic non-decreasing behavior
   - handling of wall-clock/NTP step-backs
   - one authoritative source or maxInterPepSkewSeconds for multi-PEP deployments
   - replay retention accounting for inter-PEP skew

2. Tool-call windows are named in the purpose and issue scope but are not covered normatively in configuration or enforcement. Please either add their trusted-time rules or explicitly defer them from this profile.

3. requested_ttl is used in the normative expiry formula but is not defined. Please define its source, type, bounds, trust status, and invalid-value behavior.

4. Please add basic numeric-domain requirements for evaluation_time, intent.timestamp, and all duration fields: finite/safe integer and appropriate non-negative constraints.

5. Please check the evaluate(...) result shape against the existing engine/API types. I want to avoid introducing a parallel normative result format unless the mapping is explicit.

6. Please clarify that freshness -> replay -> velocity is the ordering among trusted-time-dependent gates, not necessarily the complete global module order.

The reason-code section and the separation from clock-semantics look aligned. No verifier or runtime wiring should be added in this PR.

Whatsonyourmind:
Thanks — all six addressed in `674996a`, spec-only, no verifier or runtime wiring added.

1. **PEP time-source** → new **§2.1**: `evaluation_time` sampled exactly once per evaluation; monotonic non-decreasing within a PEP, so a wall-clock/NTP step-back can't move it backward or re-open a closed replay/velocity window; multi-PEP MUST use one authoritative clock *or* a bounded `maxInterPepSkewSeconds`; and replay retention widens to `[seen_at, seen_at + replayWindowSeconds + maxInterPepSkewSeconds)` so a nonce isn't evicted early at a lagging PEP.

2. **Tool-call windows** → explicitly **deferred** in **§9**. This profile does not make the existing `TOOL_CALL_LIMIT_EXCEEDED` path trusted-time-aware; a follow-up may, keying the window off `evaluation_time` exactly as velocity does. Flagged rather than half-specified.

3. **`requested_ttl`** → new **§5.1**: source (optional per-request input, not config), type/bounds (finite non-negative int, effective `[0, maxTtlSeconds]`), trust status (may be untrusted — safe *only* because it enters through the `min()` clamp, so it can shorten but never extend the TTL), and absent/invalid behavior (absent → `maxTtlSeconds`; invalid → fail closed, `DENY` + `STATE_INVALID`, never coerced).

4. **Numeric domain** → new **§5.2**: finite non-negative safe integers for `evaluation_time`, `intent.timestamp`, and every duration field; out-of-domain fails closed (malformed `intent.timestamp` → `DENY`/`STATE_INVALID` at the freshness gate; malformed config → refuse to evaluate), never silent coercion.

5. **Result shape** → **§3** rewritten to the existing types, no parallel format. The module returns `PolicyResult` from `packages/core/src/types/policy.ts` (`reasons`, not the invented `status`/`violations`). `issued_at`/`expiry` are **Authorization-artifact** fields (`packages/core/src/types/authorization.ts`) minted on `ALLOW` per `authorization-v1` — this profile only constrains how they're *derived* from `evaluation_time`, it doesn't add them to the module result.

6. **Ordering** → **§7** now states freshness → replay → velocity is the order **among the trusted-time gates only**, not the global module order; other `eta-core-v1` gates may interleave as long as that relative order holds.

Reason-code section left as-is since you confirmed it's aligned (still reuse-preferred, additions subject to the separate enum review). Happy to iterate further.

AngeYobo:
Thanks! the structure and lifecycle separation are strong, but I found a few normative gaps to resolve before merge.

1. Please specify the PEP-side requirements for `evaluation_time`:
   - authoritative source;
   - monotonic non-decreasing behavior, including NTP step-backs;
   - single-clock requirements or a bounded inter-PEP skew model;
   - replay retention implications when multiple PEPs are deployed.

2. The document currently mentions tool-call windows in the purpose and trust model, but does not define them in configuration, enforcement, or conformance. Please either add their normative trusted-time rules or remove them from this profile and defer them explicitly.

3. `requested_ttl` is used in the normative expiry formula but is not defined. Please define its source, type, trust boundary, and bounds, or simplify v1 to `expiry = evaluation_time + maxTtlSeconds`.

4. Please express `evaluate(intent, state, evaluation_time)` as the logical normative input model without prematurely fixing the concrete public API signature.

Minor reference corrections:
- Section 3: `derived per §5` should reference §4.
- Section 3: reason codes should reference §8, not §7.

Also please confirm the actual file does not contain the full document twice; the text shared here is duplicated.

The two-clock model, freshness ordering, replay/velocity separation, and reason-code review posture otherwise look aligned.

Whatsonyourmind:
Thanks for the re-review. I think GitHub showed you the pre-revision text — the current head is `00b9ebd` (on top of `674996a`, pushed ~13:00), and most of these already landed there. Section-by-section against the current file:

1. **PEP-side `evaluation_time`** → **§2.1**: authoritative source, monotonic non-decreasing incl. NTP/wall-clock step-backs (a backward step MUST NOT re-open a closed replay/velocity window), single authoritative clock *or* bounded `maxInterPepSkewSeconds`, and skew-aware replay retention `[seen_at, seen_at + replayWindowSeconds + maxInterPepSkewSeconds)`.
2. **Tool-call windows** → explicitly deferred in **§9** ("This profile does not make the `TOOL_CALL_LIMIT_EXCEEDED` path trusted-time-aware; a follow-up MAY…"). Removed from any normative role here.
3. **`requested_ttl`** → defined in **§5.1** (source = optional per-request input, not config; type/bounds; trust boundary — safe only via the `min()` clamp; absent → `maxTtlSeconds`; invalid → fail closed, `STATE_INVALID`). If you'd still prefer the simpler v1, I'm happy to collapse §4.3 to `expiry = evaluation_time + maxTtlSeconds` and drop `requested_ttl` entirely — your call; I kept it because the clamp makes it trust-safe, but v1 doesn't need it.
4. **Logical input model** → just pushed in `00b9ebd`: **§3** now states the evaluation is normatively a pure function of the logical inputs `(intent, state, evaluation_time)` — an *input model*, explicitly **not** a mandated public API signature; the concrete threading stays deferred (§9).

Reference corrections — both already fixed in `674996a`:
- §3 no longer says "derived per §5"; the §3 rewrite points expiry derivation to **§4** and reason codes to **§8**.
- The old `{ decision, status, issued_at?, expiry?, violations[] }` shape is gone — §3 now returns `PolicyResult` (`reasons`, no `status`/`violations`), with `issued_at`/`expiry` as **Authorization**-artifact fields, not result fields.

On the **duplication**: the file on the branch is a single copy — 281 lines, one each of §1–§10. The doubled text was in the quoted/diff view, not the source. Please pull `00b9ebd` (or view the file at the PR head) and it should read clean end to end. Happy to iterate on anything remaining.

AngeYobo:
Nice,  this is now very close.

I found two remaining normative consistency issues before merge:

1. `requested_ttl` currently allows `0`, which would produce `expiry ==
issued_at`. Please require:
- `requested_ttl` in `[1, maxTtlSeconds]`
- `maxTtlSeconds > 0`

2. §2.1 defines skew-aware replay retention for bounded-skew multi-PEP
deployments as:

`replayWindowSeconds + maxInterPepSkewSeconds`

but §7 still uses `replayWindowSeconds` alone. Please make §7 reference the
effective retention rule from §2.1 so there is only one normative
definition.

Everything else now looks aligned: the two-clock model, PEP monotonicity,
multi-PEP handling, logical input model, trusted issuance/expiry,
evaluation order, and reason-code review posture.

Once these two points are corrected, I’m comfortable merging this spec PR.

Whatsonyourmind:
Both fixed in `0e0a91b`.

1. **`requested_ttl` lower bound** — domain is now `[1, maxTtlSeconds]` (§5.1); `0` is explicitly invalid (would mint `expiry == issued_at`) → fail closed with `STATE_INVALID`. §5.2 now also requires `maxTtlSeconds ≥ 1` (and `requested_ttl ≥ 1`), so a clamp result is always `≥ 1`.

2. **Single retention definition** — the effective retention window now lives in one place, §2.1: base `[seen_at, seen_at + replayWindowSeconds)`, widened to `+ maxInterPepSkewSeconds` under bounded multi-PEP skew. §7 now references §2.1's window instead of restating `replayWindowSeconds` alone — no second definition.

Thanks for the careful passes — ready when you are.

AngeYobo:
Hi Luka, thank you so much.  Will notify you asap.

AngeYobo:
Hi Luka,

I have a few small comments before merge:

requested_ttl: I'd lean toward removing it from this first profile. While it's internally consistent (optional, bounded, fail-closed), it introduces a new input that doesn't yet exist elsewhere in ETA Core. I think v1 stays cleaner if expiry is simply:

expiry = evaluation_time + maxTtlSeconds

We can introduce per-request TTL later if we decide to support it as a separate feature.

Section references: a few references such as §4.3 / §4.4 should instead point to the corresponding invariant in §4, since §4 doesn't currently define numbered subsections.
Replay retention wording: §2.1 already defines the effective replay-retention window. I'd avoid restating the full formula again in §7 and simply reference §2.1 to keep a single normative definition.
Freshness precondition: a small cross-reference from §6 to §5.2 would make it explicit that malformed timestamps fail closed before freshness evaluation.

Regarding STATE_INVALID, the fail-closed behavior looks correct. I only wanted to confirm that this is the existing reason-code convention we already use for malformed trusted-time inputs/configuration rather than introducing a new semantic. If that's already the established convention, I'm happy with it.

Other than those points, I think the document is ready. The trusted-time model, evaluation contract, and separation of concerns are much clearer now. After these small cleanups, I'd be happy to merge it.

Whatsonyourmind:
All four done in `845f046`, plus the `STATE_INVALID` confirmation below.

1. **`requested_ttl` removed.** §4 invariant 3 is now simply `expiry = evaluation_time + maxTtlSeconds`. Dropped the whole §5.1 `requested_ttl` subsection and struck it from the §5.1 (now "Numeric Domain") duration list. Agreed it's cleaner as a later, separately-decided feature — nothing else in ETA Core carries a per-request TTL yet, so v1 shouldn't be the first.

2. **Section references.** Fixed the stale `§4.3` / `§4.4` pointers → plain `§4` (§4 has no numbered subsections). The only remaining subsection ref is §5.1, which now genuinely exists.

3. **Replay retention wording.** §7 no longer restates the formula — it points to §2.1 as the single normative definition: "over the effective retention window defined in §2.1 … this section does not restate the formula."

4. **Freshness precondition.** Added an explicit §6 → §5.1 cross-reference: a malformed `intent.timestamp` fails closed with `STATE_INVALID` *before* `Δ` is computed, so the freshness comparisons assume a well-formed timestamp.

On **`STATE_INVALID`**: it's the existing convention, not a new semantic. It's already the code every module returns for missing config / structurally corrupt state — `BudgetModule`/`ConcurrencyModule`/`RecursionDepthModule`/`ReplayModule`/`VelocityModule`/`ToolAmplificationModule` all `DENY` with `["STATE_INVALID"]` when a required field is `undefined`, and `PolicyEngine` returns it when `!isRuntimeState(state)` (the code comment there is literally "structural corruption should return STATE_INVALID (not INTERNAL_ERROR)"). Malformed trusted-time input/config is the same class, so it reuses the same code.

Head is `845f046`. Ready when you are — thanks again for the careful passes.

AngeYobo:
Done ! Thanks

PR #178, opened 2026-07-16
Add stale-intent freshness conformance vectors (#171)

Whatsonyourmind:
## Summary

Extends the trusted-time conformance vectors (`packages/conformance/vectors/trusted-time.json`) with the **stale-intent freshness direction**, which the initial draft (#174) explicitly deferred ("Stale-intent rejection … is intentionally deferred to the follow-up vector pass"). Adds three vectors mirroring the existing future-dating set into staleness, and aligns one pre-existing note with the merged spec.

New vectors:
- **`tt-neg-005`** (mirrors V1) — intent staler than `maxIntentAgeSeconds` (`−Δ > maxIntentAgeSeconds`) → `DENY`, `INTENT_STALE`, no token minted.
- **`tt-neg-006`** (mirrors V4, evaluation order) — a stale intent that is *also* a replay (retained nonce `n7`); the freshness gate runs first, so the reason is `INTENT_STALE`, not `REPLAY_NONCE`. Pins freshness-before-replay for the stale direction.
- **`tt-pos-004`** (mirrors V5, positive control) — intent inside the stale band (`−Δ ≤ maxIntentAgeSeconds`) → `ALLOW`, with `issued_at` / `expiry` derived entirely from `evaluation_time`.

## Why

The draft covered only the future-dating direction. `INTENT_STALE` (§6) is a normative freshness outcome with no coverage, and staleness is the more common real-world case (delayed or queued intents). The evaluation-order vector (`tt-neg-006`) matters because it pins that a stale intent which would also trip replay still denies on freshness first — the same ordering guarantee `tt-neg-004` fixes for future-dating.

## Boundary

- Vectors only. **No** enum change, **no** `verifyTrustedTime` wiring, **no** runtime/engine change.
- Reason codes (`INTENT_FRESHNESS_FUTURE`, `INTENT_STALE`) remain `"provisional": true`, consistent with the current spec; the enum lands in a separate change (planned sequence: enum → `verifyTrustedTime` → engine).
- These are review-stage vectors; `trusted-time.json` is not yet loaded by the conformance runner (`packages/conformance/src/validate.ts`), so there is no executable gate against them yet.

## Validation

- File parses (`json.tool`); 10 vectors, ids unique.
- Cross-checked every vector's freshness classification against `config` and spec §6 (`Δ = intent.timestamp − evaluation_time`): `tt-neg-005` `−Δ = 600 > 300` → STALE; `tt-neg-006` `−Δ = 605 > 300` → STALE (nonce retained, but freshness precedes replay); `tt-pos-004` `−Δ = 250 ≤ 300` → FRESH, `issued_at = evaluation_time`, `expiry = evaluation_time + maxTtlSeconds = T0 + 60`.
- Confirmed no `requested_ttl` / `min(maxTtlSeconds, …)` residue remains in the file.
- `pnpm -C packages/conformance validate` loads only the wired vector files (intent-hash, authorization, delegation, envelope, key-lifecycle, profile-c); `trusted-time.json` is not among them, so it is unaffected. I did not run the engine suite because these vectors don't reach it yet.

## Invariants preserved

- **Skew is freshness-only (§4.1):** the stale/future gates read `intent.timestamp` only for the freshness comparison; issuance in `tt-pos-004` still derives from `evaluation_time`.
- **Trusted issuance / bounded expiry (§4.2–4.3):** `tt-pos-004` asserts `issued_at = evaluation_time` and `expiry = evaluation_time + maxTtlSeconds`.
- **Freshness precedes replay/velocity (§7 ordering):** `tt-neg-006` pins it for the stale direction, matching `tt-neg-004` for future.
- **Fail-closed:** both new negatives mint no `issued_at` / `expiry`.

## Residual risks

- The reason-code names are provisional; if the enum settles on a single parametrized `INTENT_FRESHNESS_*`, the `INTENT_STALE` strings here need a one-line follow-up. Flagged so the provisional status stays explicit.
- No executable oracle yet — correctness rests on spec cross-reading, not a passing run. That is inherent to the pre-wiring vector stage and resolves when `verifyTrustedTime` lands and consumes this file.

## Audit/spec updates

- Updated the file-level `description` to state both freshness directions are now covered (removed the "stale … intentionally deferred" note).
- Aligned `tt-pos-001`'s `expiry_rule` string with the merged spec: `evaluation_time + maxTtlSeconds` (the prior `min(maxTtlSeconds, requested_ttl)` phrasing predated the `requested_ttl` removal in #175). No behavioral change to that vector — same `issued_at` / `expiry` values.


AngeYobo:
Thanks @Whatsonyourmind  ! the stale vectors and arithmetic look aligned.

One small consistency fix before merge: the file description still says that
`evaluation_time` governs “velocity/tool-call windows”.

The merged Trusted-Time Core Profile explicitly defers trusted-time semantics for
tool-call windows, so this draft vector file should say only:

“issuance, expiry, replay retention, and velocity windows”

No vector behavior needs to change. Once that wording is corrected, I’m comfortable merging this vectors PR.

Whatsonyourmind:
Done in `996feda` — dropped "tool-call" so the description now reads "issuance, expiry, replay retention, and velocity windows", matching the merged Core Profile's deferral of trusted-time semantics for tool-call windows. Wording-only; the 10 vectors and their arithmetic are untouched. Thanks for the careful read.

AngeYobo:
Great!

Issue #181, opened 2026-07-17
P1: Add deterministic trusted-time intent freshness verifier


AngeYobo:
### Evaluation boundary

Implement a pure deterministic trusted-time freshness verifier.

Trusted `evaluation_time` must be supplied explicitly to the verifier. The verifier must not:

- call `Date.now()`
- read process-local wall-clock state
- derive trusted time from `intent.timestamp`
- mutate replay, velocity, or authorization state

The verifier must evaluate only:

```text
intent.timestamp > evaluation_time + maxClockSkewSeconds
````

and:

```text
intent.timestamp < evaluation_time - maxIntentAgeSeconds
```

## Scope

Implement only:

* a pure trusted-time intent freshness verifier
* explicit trusted `evaluation_time` input to that verifier
* configuration handling for:

  * `maxClockSkewSeconds`
  * `maxIntentAgeSeconds`
* focused boundary and determinism tests
* public API artifacts only if required

## Non-goals

This issue must not:

* wire the verifier into `PolicyEngine`
* modify evaluation ordering
* modify replay behavior
* modify velocity behavior
* capture time at the PEP boundary
* derive authorization `issued_at`
* derive authorization `expiry`
* modify verifier behavior for existing authorizations
* activate the complete trusted-time conformance vector suite

## Acceptance criteria

* [ ] trusted `evaluation_time` is supplied explicitly
* [ ] no hidden wall-clock access is introduced
* [ ] future-dated intent returns `INTENT_FRESHNESS_FUTURE`
* [ ] stale intent returns `INTENT_STALE`
* [ ] both exact boundaries are accepted
* [ ] repeated evaluation with identical inputs produces identical output
* [ ] replay and velocity code are untouched
* [ ] PolicyEngine behavior is unchanged
* [ ] public API impact is documented
* [ ] full validation passes

Issue #184, opened 2026-07-17
P1: Integrate trusted-time freshness before replay and velocity gates


AngeYobo:
## Summary

Integrate the merged deterministic `verifyTrustedTime` primitive into the approved policy evaluation boundary.

The freshness gate must execute before replay and velocity checks so that attacker-controlled timestamps cannot reach later stateful gates before trusted-time freshness has been established.

## Security invariant

A request with an invalid freshness claim must be denied before replay or velocity evaluation.

Evaluation order:

```text
trusted input validation
→ intent timestamp validation
→ trusted-time freshness
→ replay
→ velocity
→ remaining policy evaluation
````

A thrown trusted-time precondition failure must fail closed at the engine boundary.

It must never be converted into `ALLOW` or silently ignored.

### Future-dated request

When:

```text
intent.timestamp > evaluation_time + maxClockSkewSeconds
```

return:

```text
DENY / INTENT_FRESHNESS_FUTURE
```

before replay or velocity evaluation.

### Stale request

When:

```text
intent.timestamp < evaluation_time - maxIntentAgeSeconds
```

return:

```text
DENY / INTENT_STALE
```

before replay or velocity evaluation.

### Fresh request

When the timestamp is inside the inclusive freshness band, evaluation continues into the existing replay and velocity gates without changing their current semantics.

### Trusted-input failure

If `verifyTrustedTime` throws because trusted `evaluation_time` or trusted configuration is malformed, the engine boundary must abort and fail closed.

The exact existing engine-level denial/error representation must be selected consistently with current repository conventions. Do not invent a parallel result type.

A trusted-precondition exception must not be converted into `STATE_INVALID` unless an existing, explicitly applicable engine convention already requires that mapping.

The implementation must not attribute malformed trusted time or trusted configuration to the attacker-controlled intent.

## Scope

Implement only:

* threading explicit `evaluation_time` into the smallest approved policy evaluation boundary
* invoking `verifyTrustedTime` before replay and velocity
* fail-closed handling of trusted-time precondition exceptions
* focused evaluation-order tests
* directly required API artifacts

## Required tests

Cover at minimum:

1. fresh intent continues through normal evaluation
2. future-dated intent denies with `INTENT_FRESHNESS_FUTURE`
3. stale intent denies with `INTENT_STALE`
4. future-dated intent with retained nonce still returns `INTENT_FRESHNESS_FUTURE`
5. stale intent with retained nonce still returns `INTENT_STALE`
[...]

PR #185, opened 2026-07-18
test(core): harden trusted-time verifier boundary invariants (#183)

### LINKED ISSUE #183: P1: Harden trusted-time verifier boundary invariants
## Summary

Add focused regression coverage for two security-relevant invariants already implemented by `verifyTrustedTime`:

1. overflow-safe freshness comparison near `Number.MAX_SAFE_INTEGER`
2. trusted-input validation ordering before attacker-controlled input validation

This issue is test-only and must not change runtime behavior.

## Why

`verifyTrustedTime` currently evaluates freshness using difference form:

```text
delta = intentTimestamp - evaluationTime
````

rather than computing:

```text
evaluationTime + maxClockSkewSeconds
```

The difference form avoids precision loss when trusted timestamps are near JavaScript's maximum safe integer.

The implementation also validates trusted caller inputs before validating attacker-controlled `intentTimestamp`. This ensures a broken trusted clock or invalid configuration surfaces as a precondition failure rather than being masked as `DENY / STATE_INVALID`.

Both properties should be pinned by regression tests before the verifier is wired into `PolicyEngine`.

### Maximum-magnitude arithmetic

Add focused cases near:

```text
Number.MAX_SAFE_INTEGER
```

that prove:

* a timestamp at the accepted boundary remains accepted
* a timestamp beyond the permitted boundary denies correctly
* the implementation does not rely on unsafe addition such as:

  * `evaluationTime + maxClockSkewSeconds`

Use only numerically valid safe integers.

### Validation ordering

Add a case where:

* at least one trusted input is malformed
* `intentTimestamp` is also malformed

The result must be a thrown trusted-precondition failure, not:

```text
DENY / STATE_INVALID
```

This pins the ordering:

```text
trusted inputs
→ attacker-controlled intent timestamp
→ freshness evaluation
```

## Scope

Modify only the focused trusted-time verifier test file unless a test exposes an actual implementation defect.

Expected file:

```text
packages/core/src/test/verify.trusted-time.test.ts
```

## Non-goals

This issue must not:

* modify `PolicyEngine`
* modify `verifyTrustedTime` behavior
* export `verifyTrustedTime` publicly
* modify replay logic
* modify velocity logic
* modify PEP clock capture
* modify authorization issuance or expiry
* activate trusted-time conformance vectors
* modify the Trusted-Time Core Profile
* refactor unrelated code

## Acceptance criteria

* [ ] maximum safe-integer boundary behavior is covered
* [ ] unsafe sum-form regression is detectable
* [ ] trusted-input validation precedence is covered
* [ ] no production behavior changes
* [ ] no public API changes
* [ ] full validation passes

## Validation

```bash
git diff --check
pnpm --filter @oxdeai/core typecheck
pnpm --filter @oxdeai/core test
pnpm typecheck
pnpm test
pnpm test:vectors:all
pnpm security:gate
```

## Final report

Report:

* files changed
* tests added
* exact maximum-magnitude values used
* validation-order case added
* confirmation that production code is unchanged
* API fingerprint impact
* validation results

Refs #181
Refs #171

Whatsonyourmind:
## Summary

Adds focused, test-only regression coverage for two security-relevant invariants already implemented by `verifyTrustedTime`, per #183:

1. **Overflow-safe freshness comparison** near `Number.MAX_SAFE_INTEGER` (difference form, no unsafe addition).
2. **Trusted-input validation ordering** — trusted preconditions are validated before the attacker-controlled `intentTimestamp`.

Only `packages/core/src/test/verify.trusted-time.test.ts` changes. No production code, no public API, no behavior change.

## Why

`verifyTrustedTime` evaluates freshness as `delta = intentTimestamp - evaluationTime` and compares `delta` to the tolerances, rather than the literal spec sum form `intentTimestamp > evaluationTime + maxClockSkewSeconds`. Since `isProtocolSeconds` bounds every operand to `[0, MAX_SAFE_INTEGER]`, `delta` is always an exact safe integer, while `evaluationTime + maxClockSkewSeconds` can exceed 2^53 and lose integer precision. It also validates the trusted caller inputs (`evaluationTime`, config bounds) before the attacker-controlled `intentTimestamp`, so a broken trusted clock or invalid configuration surfaces as a thrown precondition failure rather than being masked as `DENY / STATE_INVALID`. Both properties are currently implicit; these tests pin them before the verifier is wired into `PolicyEngine`.

## Boundary (non-goals honored)

No change to `verifyTrustedTime` behavior, `PolicyEngine`, public exports, replay/velocity logic, PEP clock capture, issuance/expiry, conformance-vector activation, or the Trusted-Time Core Profile. No unrelated refactor. Test file only.

## Tests added (7; core suite 212 → 219)

**Maximum-magnitude arithmetic** (`MAX = Number.MAX_SAFE_INTEGER = 9007199254740991`):
- `evaluationTime = MAX - 300`, `skew = 300`, `intent = MAX` → **ALLOW** (future boundary at exactly `MAX`, inclusive).
- `evaluationTime = MAX - 300`, `skew = 299`, `intent = MAX` → **INTENT_FRESHNESS_FUTURE** (`delta = 300 > 299`, all valid safe integers).
- `evaluationTime = 1000`, `skew = MAX - 1`, `intent = MAX` → **ALLOW**. Here `evaluationTime + maxClockSkewSeconds = MAX + 999` (beyond the safe range), yet the difference form keeps `delta` exact and correctly sees the intent as within skew.
- `evaluationTime = MAX`, `maxIntentAge = 300`: `intent = MAX - 300` → **ALLOW** (stale boundary), `intent = MAX - 301` → **INTENT_STALE**.
- `intent = MAX + 1` (= 2^53, not a safe integer) → **DENY STATE_INVALID** (out-of-domain data, never evaluated for freshness).

**Validation ordering**:
- malformed `evaluationTime` (`NaN`) **and** malformed `intentTimestamp` (`NaN`) → **throws** `/evaluationTime/` (trusted precondition precedes the `STATE_INVALID` data path).
- malformed config (`maxClockSkewSeconds = NaN`) **and** malformed `intentTimestamp` (`Infinity`) → **throws** `/maxClockSkewSeconds/`.

The ordering tests are the strong pins: if validation were reordered to check `intentTimestamp` first, both would return `DENY / STATE_INVALID` instead of throwing, and the assertions would fail.

## Honest scope note on "sum-form regression detectable"

Worth stating precisely, because it bears on acceptance criterion 2. Given the `isProtocolSeconds` domain guard (both operands ≤ `MAX_SAFE_INTEGER`), the difference form and the plain sum form are **behaviorally equivalent on every valid input**: wherever `evaluationTime + maxClockSkewSeconds` would overflow the safe range, no valid `intentTimestamp` (≤ `MAX_SAFE_INTEGER`) can reach that boundary, so both forms `ALLOW`; wherever the sum is within range it is exact. So a black-box test cannot distinguish a plain sum-form rewrite from the difference form on valid inputs — the difference form's benefit is the **absence of an unsafe intermediate**, not a divergent valid-input result.

What these max-magnitude tests concretely guarantee: correct behavior at the safe-integer domain edge (future/stale boundaries and out-of-domain rejection), and — via the `evaluationTime = 1000, skew = MAX - 1` case — that a sum-form rewrite which *guards or throws* on the unsafe `evaluationTime + maxClockSkewSeconds` intermediate would regress and be caught. If you'd like a stronger guarantee against the plain-sum rewrite specifically, that needs a source-level assertion rather than a black-box test (out of this issue's "test file only" scope) — happy to follow up if you want it.

## Protocol / security

- **Invariants preserved:** deterministic ALLOW/DENY, inclusive boundaries via strict `>`, fail-closed on malformed data (`STATE_INVALID`) vs throw on malformed trusted precondition, purity/immutability. All unchanged — this PR only adds assertions over them.
- **Residual risks:** none introduced (test-only). The sum-form-detectability limit above is documented, not newly created.
- **Audit-spec updates:** none. No Core Profile or vector changes.

## Validation

```
git diff --check                          → clean
pnpm --filter @oxdeai/core typecheck      → PASS (tsc --noEmit, exit 0)
pnpm --filter @oxdeai/core test           → PASS (build + node --test: 219 pass / 0 fail)
git diff -- packages/core/API_FINGERPRINT → no change (test-only)
git status --short                        → only packages/core/src/test/verify.trusted-time.test.ts
```

Full `pnpm typecheck` / `pnpm test` / `pnpm -C packages/conformance validate` / `pnpm test:vectors:all` / `pnpm security:gate` run in CI (the vector harnesses need the Go/Python toolchains); this change touches no production or vector surface.

## Final report

- **Files changed:** `packages/core/src/test/verify.trusted-time.test.ts` (only).
- **Tests added:** 7 — 5 maximum-magnitude arithmetic (one asserts both the stale boundary and one-second-beyond in a single case) + 2 validation-ordering. Core suite 212 → 219.
- **Exact maximum-magnitude values used:** `MAX = 9007199254740991`; `MAX - 300 = 9007199254740691`; `MAX - 301`; `MAX - 1 = 9007199254740990`; `MAX + 1 = 9007199254740992` (= 2^53, unsafe); `evaluationTime = 1000` with `skew = MAX - 1` (sum `= MAX + 999`, beyond safe range).
- **Validation-order case added:** malformed-`evaluationTime` + malformed-`intentTimestamp` → throws `/evaluationTime/`; malformed-config + malformed-`intentTimestamp` → throws `/maxClockSkewSeconds/`.
- **Production code unchanged:** yes — test file only.
- **API fingerprint impact:** none.
- **Validation results:** typecheck PASS (exit 0); `@oxdeai/core` test 219 pass / 0 fail; whitespace clean; API fingerprint unchanged; only the test file modified (see Validation section above).

Refs #171, #181. Closes #183.

### DIFF
--- a/packages/core/src/test/verify.trusted-time.test.ts
+++ b/packages/core/src/test/verify.trusted-time.test.ts
@@ -169,3 +169,83 @@ test("verifyTrustedTime: input object remains unchanged after verification", () 
   assert.deepEqual(input, snapshot);
   assert.equal(out.decision, "DENY");
 });
+
+// ── Maximum-magnitude arithmetic (overflow-safe freshness, issue #183) ───────
+//
+// Freshness is evaluated in DIFFERENCE form (`delta = intentTimestamp -
+// evaluationTime`, compared to the tolerances), not the literal spec SUM form
+// (`intentTimestamp > evaluationTime + maxClockSkewSeconds`). Because
+// `isProtocolSeconds` bounds every operand to `[0, MAX_SAFE_INTEGER]`, `delta`
+// is always an exact safe integer, whereas `evaluationTime +
+// maxClockSkewSeconds` can exceed 2^53 and lose integer precision. These cases
+// pin correct behavior at the safe-integer domain edge and exercise the path
+// where a naive sum would form an unsafe intermediate.
+
+const MAX = Number.MAX_SAFE_INTEGER;
+
+test("verifyTrustedTime: intent exactly at the future boundary with evaluationTime near MAX_SAFE_INTEGER → ALLOW", () => {
+  // Future boundary = evaluationTime + skew = MAX (exactly representable); inclusive → ALLOW.
+  const out = verifyTrustedTime(
+    baseInput({ intentTimestamp: MAX, evaluationTime: MAX - 300, maxClockSkewSeconds: 300 })
+  );
+  assert.deepEqual(out, { decision: "ALLOW", reasons: [] });
+});
+
+test("verifyTrustedTime: intent one second beyond the future boundary near MAX_SAFE_INTEGER → INTENT_FRESHNESS_FUTURE", () => {
+  // delta = 300, skew = 299 → strictly beyond → deny; all operands are valid safe integers.
+  const out = verifyTrustedTime(
+    baseInput({ intentTimestamp: MAX, evaluationTime: MAX - 300, maxClockSkewSeconds: 299 })
+  );
+  assert.deepEqual(out, { decision: "DENY", reasons: ["INTENT_FRESHNESS_FUTURE"] });
+});
+
+test("verifyTrustedTime: fresh intent when evaluationTime + maxClockSkewSeconds would exceed MAX_SAFE_INTEGER → ALLOW (no unsafe addition)", () => {
+  // evaluationTime + maxClockSkewSeconds = MAX + 999 (beyond the safe range), yet the difference
+  // form keeps delta exact, so the intent is correctly within skew. A sum-form implementation that
+  // guarded or threw on the unsafe intermediate would regress this case.
+  const out = verifyTrustedTime(
+    baseInput({ intentTimestamp: MAX, evaluationTime: 1000, maxClockSkewSeconds: MAX - 1 })
+  );
+  assert.deepEqual(out, { decision: "ALLOW", reasons: [] });
+});
+
+test("verifyTrustedTime: stale boundary is exact at maximum magnitude → ALLOW at the boundary, INTENT_STALE beyond", () => {
+  const atBoundary = verifyTrustedTime(
+    baseInput({ intentTimestamp: MAX - 300, evaluationTime: MAX, maxIntentAgeSeconds: 300 })
+  );
+  assert.deepEqual(atBoundary, { decision: "ALLOW", reasons: [] });
+  const beyond = verifyTrustedTime(
+    baseInput({ intentTimestamp: MAX - 301, evaluationTime: MAX, maxIntentAgeSeconds: 300 })
+  );
+  assert.deepEqual(beyond, { decision: "DENY", reasons: ["INTENT_STALE"] });
+});
+
+test("verifyTrustedTime: intent one past MAX_SAFE_INTEGER is out of domain → DENY STATE_INVALID (not a freshness code)", () => {
+  // MAX + 1 (= 2^53) is not a safe integer, so it is rejected as malformed data, never evaluated for freshness.
+  const out = verifyTrustedTime(
+    baseInput({ intentTimestamp: MAX + 1, evaluationTime: MAX - 300, maxClockSkewSeconds: 300 })
+  );
+  assert.deepEqual(out, { decision: "DENY", reasons: ["STATE_INVALID"] });
+});
+
+// ── Validation ordering: trusted preconditions before attacker input (issue #183) ──
+//
+// Trusted inputs (evaluationTime, config bounds) are validated BEFORE the
+// attacker-controlled intentTimestamp. When BOTH a trusted input and the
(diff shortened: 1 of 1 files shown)


AngeYobo:
Thanks Luka, reviewed.
This is exactly the focused hardening we wanted before the PolicyEngine integration:

* maximum safe-integer boundary coverage;
* explicit protection against unsafe-intermediate handling;
* trusted-input validation ordering pinned before attacker-controlled timestamp validation;
* test-only, with no production or API change.

I also agree with the precise note on black-box detectability of a plain sum-form rewrite. 
The important invariant here is avoiding unsafe intermediate arithmetic, and the new cases lock the safe-domain behavior without expanding the scope into source-level assertions.
Approved to merge. After this lands, we’ll proceed with the separately scoped #184 PolicyEngine integration.

PR #196, opened 2026-07-21
fix(types): reconcile AuthorizationV1 auth_id reads with legacy authorization_id (#186)

### LINKED ISSUE #186: P2: Reconcile AuthorizationV1 auth_id with legacy authorization_id alias
## Summary

`bench/cases/evaluate.ts` accesses:

```ts
result.authorization.authorization_id
```

However, `PolicyEngine.evaluatePure()` exposes the ALLOW authorization as `AuthorizationV1`, whose normative identifier field is:

```ts
auth_id
```

`AuthorizationV1` does not declare the legacy alias:

```ts
authorization_id
```

The runtime authorization object currently appears to carry both fields through the broader `Authorization = AuthorizationLegacy & AuthorizationV1` construction, but the compile-time type at this call site is the narrower `AuthorizationV1`.

## Observed error

```text
Property 'authorization_id' does not exist on type 'AuthorizationV1'. ts(2339)
```

## Background

This mismatch was discovered during the implementation of trusted-time integration (#184).

The investigation indicates that this is a pre-existing type/runtime inconsistency rather than a regression introduced by #184.

No fix was applied in #184 to keep the trusted-time change focused and avoid expanding its scope.

## Impact

- Benchmark code relies on a legacy runtime alias not represented by the public TypeScript contract.
- Runtime behavior and compile-time types are inconsistent.
- Future cleanup of legacy authorization fields could silently break benchmark code.
- The canonical public authorization identifier is currently ambiguous.

## Decision required

This issue must explicitly determine which identifier is the supported public contract.

### Option A : Migrate callers to `auth_id` (preferred)

Treat `auth_id` as the canonical identifier and migrate benchmark and any remaining internal callers away from `authorization_id`.

This is the preferred direction if `authorization_id` is only a compatibility alias and is **not** part of the supported public API.

### Option B : Widen the public authorization type

Change the relevant engine return type so that `authorization_id` is formally declared alongside `auth_id`.

This is only appropriate if the legacy alias is intentionally part of the public compatibility contract.

### Option C : Deprecate then remove the alias

Document `authorization_id` as deprecated, migrate all internal callers to `auth_id`, and remove the alias in a later semver-major release.

## Questions to resolve

1. Is `authorization_id` part of the supported public API?
2. Is it only a runtime compatibility alias?
3. Should all internal code migrate to `auth_id`?
4. Are there additional call sites still relying on `authorization_id`?
5. Should the public return type of `evaluatePure()` change?

## Recommended direction

Unless repository specifications, published API documentation, or compatibility guarantees explicitly require `authorization_id`, the preferred resolution is:

- keep `AuthorizationV1` as the normative public type;
- migrate benchmark and remaining internal callers to `auth_id`;
- avoid widening `evaluatePure()` to a legacy combined type solely to satisfy benchmark code.

A public API should not be broadened merely to preserve an internal legacy alias.

## Acceptance criteria

- Determine whether `authorization_id` is an intentional public compatibility field or only a legacy runtime alias.
- Document the canonical public authorization identifier.
- Migrate all affected internal callers to the selected contract.
- Align runtime behavior and TypeScript types.
- Add a compile-time regression test.
- Update API documentation if the public type changes.
- Avoid changing authorization hashes, signatures, canonicalization, serialization, or wire format.
- Explicitly classify the change as semver-patch, semver-minor, or semver-major.

## Scope

This issue is intentionally separate from #184.

No trusted-time behavior, replay logic, velocity logic, authorization issuance, canonicalization, or security semantics should change as part of this work.

Whatsonyourmind:
Closes #186.

Per your confirmation on the issue: **Option A, scoped strictly to role 1** (artifact-identifier reads). Patch-level internal consistency fix — no wire-format, canonicalization, serialization, or API change.

## What changed

The public engine surface — the `ALLOW` branch of `EvaluateOutput` / `EvaluatePureOutput`, and `AuthorizationV1` itself — exposes the canonical `auth_id` and not the legacy `authorization_id`, so reading `.authorization_id` off those public values is a `TS2339` (the exact error in the issue). Migrated the identified reads:

| Site | Change |
|------|--------|
| `bench/cases/evaluate.ts:41` | `out.authorization.authorization_id` → `auth_id` (the reported TS2339) |
| `examples/gpu-guard/demo.mjs:58` | read → `auth_id` (was `undefined` at runtime on the public output); console label updated to match |
| `packages/core/README.md:510` | RELEASE-intent example **value** read → `out.authorization.auth_id` |
| `examples/delegation-demo/src/scenario.ts:64` | doc-comment consistency only |

**Left unchanged, deliberately:**
- Protocol **reference** fields — `authorization_id` in RELEASE intents (including the README example's `authorization_id:` **key**), audit events, conformance vectors, canonicalization/hashing.
- CLI / output payload keys.

## Compile-time regression guard

Added `packages/core/src/test/authorization-v1-boundary.compile.test.ts`. It uses conditional-type assertions (`"authorization_id" extends keyof <public type> ? "LEAK" : "ok"`) that stop type-checking if `authorization_id` re-enters `AuthorizationV1` or a public `ALLOW` authorization is re-widened back to the legacy-carrying `Authorization` type. Because the core build runs with `noEmitOnError`, that regression fails both `pnpm --filter @oxdeai/core typecheck` and the build — turning a silent boundary drift into a hard error. The existing `public-artifact-boundary.test.ts` continues to cover the *runtime* stripping via `toPublicAuthorizationV1()`; this file guards the *static type* contract.

## Verification

- `pnpm --filter @oxdeai/core typecheck` — clean
- `pnpm --filter @oxdeai/core test` — **255/255 pass** (incl. the new guard)
- `bench` typecheck (against freshly-built core) — clean **except one pre-existing, unrelated error** (see below)

## Two things for your call (not fixed here, to respect the confirmed scope)

1. **`examples/delegation-demo/src/scenario.ts:64`** is a doc comment, not a read — I updated it for naming consistency since it sits directly on the `authId` field, but happy to drop it if you'd rather keep this PR to reads only.
2. **`bench/fixtures.ts:192`** has a pre-existing `AuthorizationV1` → `Authorization` widening error (the fixture's `auth` field is typed as the full `Authorization` but receives the public `out.authorization`). It's untouched by this PR and is the *inverse* direction of the role-1 reads, so I left it out. If you want, I can reconcile it in a follow-up (retype the fixture field to `AuthorizationV1`).

Signed-off-by is on the commit (DCO). Let me know if you'd like either scope tweak.

### DIFF
--- a/bench/cases/evaluate.ts
+++ b/bench/cases/evaluate.ts
@@ -38,7 +38,7 @@ export function create(seed: number): () => unknown {
   return () => {
     const out = engine.evaluatePure(fx.intent, cloneState(baseState), fx.intent.timestamp);
     if (out.decision === "ALLOW") {
-      return out.authorization.authorization_id;
+      return out.authorization.auth_id;
     }
     return out.decision;
   };
--- a/examples/delegation-demo/src/scenario.ts
+++ b/examples/delegation-demo/src/scenario.ts
@@ -61,7 +61,7 @@ export interface AuthDecision {
   authType: "engine" | "delegation";
   tool: string;
   args: Record<string, string | number>;
-  /** For engine auth: the authorization_id */
+  /** For engine auth: the auth_id */
   authId?: string;
   /** For delegation: the delegation_id */
   delegationId?: string;
--- a/examples/gpu-guard/demo.mjs
+++ b/examples/gpu-guard/demo.mjs
@@ -55,7 +55,7 @@ const out = engine.evaluatePure(intent, state, { mode: "fail-fast" });
 
 console.log("decision:", out.decision);
 if (out.decision === "ALLOW") {
-  console.log("authorization_id:", out.authorization.authorization_id);
+  console.log("auth_id:", out.authorization.auth_id);
   console.log("policyId:", engine.computePolicyId());
   console.log("stateHash:", engine.computeStateHash(out.nextState));
   console.log("auditHeadHash:", engine.audit.headHash());
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -507,7 +507,7 @@ const releaseIntent: Intent = {
   metadata_hash: "0".repeat(64),
   signature: "",
   type: "RELEASE",
-  authorization_id: out.authorization.authorization_id,
+  authorization_id: out.authorization.auth_id,
   nonce: 43n,
   amount: 0n,
   timestamp: now
--- a/packages/core/src/test/authorization-v1-boundary.compile.test.ts
+++ b/packages/core/src/test/authorization-v1-boundary.compile.test.ts
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * authorization-v1-boundary.compile.test.ts
+ *
+ * Compile-time regression guard for the public AuthorizationV1 artifact
+ * boundary (P0-4, resolved in #102 / #103; reconciliation tracked in #186).
+ *
+ * The public engine surface — `AuthorizationV1` and the ALLOW branch of
+ * `EvaluateOutput` / `EvaluatePureOutput` — exposes the canonical `auth_id`
+ * identifier and MUST NOT statically expose the legacy `authorization_id`
+ * alias. Reading `.authorization_id` off any of these public types is exactly
+ * the TS2339 that motivated migrating the artifact-identifier reads to
+ * `auth_id`.
+ *
+ * The assertions below are conditional types. If a future change re-widens a
+ * public authorization back to the legacy-carrying `Authorization` type (or
+ * re-adds `authorization_id` to `AuthorizationV1`), the guarded type flips to
+ * "LEAK" and the corresponding `const` initializer stops type-checking, so
+ * `tsc` fails under both `pnpm --filter @oxdeai/core typecheck` and the build
+ * (`noEmitOnError`). This turns a silent boundary regression into a hard error.
+ *
+ * Scope note: legacy fields are still physically present on the runtime object
+ * by design; that is covered separately by public-artifact-boundary.test.ts,
+ * which proves `toPublicAuthorizationV1()` strips them from every signing/hash
+ * surface. This file guards the *static type* contract only.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import type { AuthorizationV1, EvaluateOutput, EvaluatePureOutput } from "@oxdeai/core";
+
+// The `authorization` payload carried by each public ALLOW result.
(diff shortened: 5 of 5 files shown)


Whatsonyourmind:
CI note: the failing `security` check is the dependency-advisory gate (`security-gate.mjs` over `pnpm audit`), not this PR. It's blocking on two newly-published advisories against pre-existing transitive **dev** dependencies:

- `brace-expansion` (high, id=1123898) via `@microsoft/api-extractor → minimatch`
- `js-yaml` (high, id=1123911) via `puppeteer → cosmiconfig`

Both versions are already in `main`'s lockfile, and `main`'s own `security-gate` was green on 2026-07-20 with that same lockfile — so these advisories landed since then and now trip the gate on any open PR (and presumably the next push to `main`). This PR changes no dependencies (source reads + one test file), so it neither introduces nor can resolve the finding. The other three checks — `build-test`, `adapter-validation`, `pep-redis-replay` — are green.

Happy to open a separate, scoped PR to clear it whichever way you prefer — a `pnpm.overrides` bump of the two transitive deps, or dated exceptions in `security/vuln-policy.json` — and keep this one focused on #186.

AngeYobo:
Thanks. 
The implementation matches the confirmed scope and the three-role distinction.
Please keep the delegation-demo comment update; it improves terminology consistency without changing behavior.
Leave bench/fixtures.ts unchanged in this PR. The AuthorizationV1 -> Authorization widening is a separate type-boundary cleanup and should not expand #186.
The compile-time regression guard is appropriate, and the runtime boundary remains covered separately.
Regarding CI, please keep the dependency-advisory remediation out of this PR. I’ll track and resolve the newly published advisories independently, then we can rebase this branch and rerun the complete checks.

No other scope changes requested.

Whatsonyourmind:
Understood — I'll leave the branch exactly as-is: delegation-demo comment kept, `bench/fixtures.ts` untouched (agreed it's a separate type-boundary cleanup, not part of #186), and no dependency changes here. The branch is ready to rebase onto `main` and rerun the full checks whenever you've resolved the advisories — happy to do the rebase on your signal, or for you to rebase directly, whichever is easier. Thanks for the review.

AngeYobo:
Hi Luka, #198 is merged. Could you please rebase #196 onto main, rerun CI/security, and push? Once it's green I'll merge it. Thanks!

Whatsonyourmind:
Done — rebased onto `main` (now includes #198 and #199) and force-pushed. Verified locally on the rebased branch: core typecheck clean, `@oxdeai/core` tests 271/271 green, and the security gate now reports `Decision: ALLOW` (0 blocking findings) thanks to #198. CI is re-running on the new head now. Thanks for sorting the advisories!

Issue #173, opened 2026-07-09
P1: Clarify or harden engine.verifyAuthorization HMAC verification surface


AngeYobo:
## Summary

Clarify or harden the `PolicyEngine.verifyAuthorization` HMAC verification surface.

A local verification-surface review found that `engine.verifyAuthorization` validates the engine HMAC over only a subset of the authorization fields. Mutating some fields such as `audience`, `capability`, or `state_hash` can still result in `engine.verifyAuthorization` returning valid.

The strict standalone verifier rejects the same mutation via the full authorization signature path. Therefore, this is not a break of the strict verification protocol. It is a dangerous API-surface ambiguity: an integrator may treat `engine.verifyAuthorization` as full authorization verification when it only authenticates the engine-specific HMAC subset.

## Problem

The engine HMAC appears to cover a subset similar to:

```text
intent_hash
policy_version
state_snapshot_hash
decision
expires_at
````

Fields outside that subset may be mutable without invalidating `engine.verifyAuthorization`.

Examples of sensitive fields that should not be ambiguous:

```text
audience
capability
state_hash
```

If an integrator uses `engine.verifyAuthorization` as the enforcement gate instead of the strict verifier, these fields may be treated as authenticated when they are not.

## Goal

Either harden `engine.verifyAuthorization` so it authenticates the full authorization core, or make its limited scope explicit and difficult to misuse.

Preferred options:

1. Expand the engine HMAC coverage to include the same security-relevant fields as the authorization core.
2. Rename or mark the method as internal/legacy/non-strict if it is not intended for relying-party enforcement.
3. Return metadata indicating the verification surface.
4. Route callers toward the strict verifier for external enforcement decisions.

## Security posture

Strict Ed25519 verification rejects the tested mutations.

This issue concerns API ambiguity and HMAC verification coverage in the engine helper method.

## Suggested investigation

Inspect:

```text
packages/core/src/policy/PolicyEngine.ts
```

and related verification helpers.

Determine:

* exact fields covered by `engine_signature`
* intended use of `engine.verifyAuthorization`
* whether relying parties are expected to call it
* whether `state_hash`, `audience`, and `capability` should be covered
* whether this method should be public API

## Tests to add

Add tests covering:

* mutation of `audience`
* mutation of `capability`
* mutation of `state_hash`
* strict verifier rejects those mutations
* `engine.verifyAuthorization` either rejects them or clearly reports limited verification
* documentation/comments prevent misuse as full verification

## Non-goals

[...]

PR #201, opened 2026-07-23
feat(verification): expose signature verification status and coverage on authorization results (#172)

### LINKED ISSUE #172: P1: Harden verification defaults and expose signature verification status
## Summary

Harden verification defaults and make signature verification status explicit in verification results.

A local verification-surface review found that calling the verification API without strict options can return `ok: true` even when cryptographic signature verification has not actually been performed.

Strict verification rejects forged signatures correctly. The issue is the permissive default API behavior, which creates an integration footgun.

## Problem

Current risky pattern:

```ts
verifyAuthorization(auth)
````

If called without:

```ts
mode: "strict"
trustedKeySets
requireSignatureVerification
```

the verification path may skip signature verification and still return:

```text
ok: true
status: ok
```

This can mislead integrators into believing an authorization was cryptographically verified when it was only structurally or semantically checked.

## Goal

Make it impossible or obvious when an authorization result has not been cryptographically verified.

Possible acceptable approaches:

1. Make signature verification required by default.
2. Keep permissive mode but expose `signatureVerified: boolean`.
3. Return a distinct status when signature verification is skipped.
4. Require explicit opt-in for non-cryptographic validation mode.

The preferred approach should preserve compatibility where necessary while making unsafe usage visible.

## Security posture

Strict Ed25519 verification already rejects forged signatures.

This issue is not claiming a protocol signature break. It is an API safety issue: the default or easy path should not look equivalent to full verification.

## Suggested approach

Review verification code paths and result types.

Add one or more of:

```ts
signatureVerified: boolean
verificationMode: "strict" | "permissive" | "structure-only"
```

Consider changing default behavior so signature verification is required unless explicitly disabled.

If compatibility requires permissive default behavior, the result must clearly indicate that no signature verification occurred.

## Tests to add

Add tests covering:

* forged signature rejected in strict mode
* forged signature not reported as fully verified in default/permissive mode
* `signatureVerified === true` only when cryptographic verification actually ran and succeeded
* missing trusted key set does not silently produce a fully verified result
* explicit non-crypto validation mode is distinguishable from strict verification

## Non-goals

This issue must not:

* change authorization artifact canonicalization
* change Ed25519 signature semantics
* change `AuthorizationV1` unless explicitly required and justified
* weaken strict verification
* remove strict verification
* silently accept forged signatures as verified
* mix with `engine.verifyAuthorization` HMAC surface hardening

## Acceptance criteria

* [ ] verification result indicates whether signature verification occurred
* [ ] forged signatures are rejected in strict mode
* [ ] default/permissive verification cannot be mistaken for full cryptographic verification
* [ ] tests cover strict and non-strict verification paths
* [ ] documentation or comments clarify safe usage
* [ ] no protocol semantics changed unless explicitly justified
* [ ] API fingerprint impact is reviewed and documented
* [ ] validation passes

## Validation

Run:

```bash
git diff --check
pnpm typecheck
pnpm test
pnpm test:vectors:all
pnpm security:gate
```

Expected:

* typecheck passes
* tests pass
* conformance passes
* vectors pass
* security gate passes
* API fingerprint change documented if public types change

## Risk

This touches verification API semantics. Keep it separate from trusted-time fixes and engine HMAC verification hardening.

Whatsonyourmind:
Closes #172.

## What

`verifyAuthorization` could return `ok: true` on the permissive default path (no `trustedKeySets`, `requireSignatureVerification` unset) **without any cryptographic signature verification having run** — the classic integration footgun called out in #172. This makes the fact of verification explicit and unambiguous on the result, without changing default behavior.

Per the discussion on #172, the three facts are kept as **orthogonal fields** (rather than overloading one enum):

```ts
signatureVerified: boolean;                                   // was a signature actually checked?
verificationMode: "strict" | "permissive" | "structure-only"; // HOW verification was performed
verificationCoverage: "authorization-v1-full" | "engine-hmac-subset" | "none"; // WHAT surface was covered
```

- `signatureVerified` flips `true` **only** at the point a cryptographic check actually runs *and* succeeds.
- `verificationMode` describes the posture actually applied (distinct from the `mode: "strict" | "best-effort"` request option): `"structure-only"` when no crypto check was engaged, `"permissive"` when one ran in best-effort, `"strict"` when strict was requested.
- `verificationCoverage` records the covered surface. This verifier only ever emits `"authorization-v1-full"` or `"none"`; the `"engine-hmac-subset"` value is defined here so **#173 can reuse the same vocabulary** on `PolicyEngine.verifyAuthorization` without introducing a parallel status model.

## Types

- `verifyAuthorization` now returns a new `AuthorizationVerificationResult` — a `VerificationResult` in which the three descriptors are **required** (always populated), so an authorization result can never be silently mistaken for verified.
- The shared `VerificationResult` gains the three fields as **optional** (additive, non-breaking for the other verifiers), carrying the shared vocabulary for #173.
- New exported types: `VerificationMode`, `VerificationCoverage`, `AuthorizationVerificationResult`.

## Orthogonal to `ok`

A validly-signed but **expired** authorization is now `signatureVerified: true`, `ok: false` — the signature check and overall validity are reported independently.

## Compatibility

Default/permissive behavior is **unchanged**: a structurally-valid authorization with no trust anchor is still `ok: true`. The result simply now also says `signatureVerified: false`, `verificationMode: "structure-only"`, `verificationCoverage: "none"`, so a caller inspecting the result cannot be misled. Making strict verification the default is intentionally left as a separate, fingerprint-reviewed follow-up.

## Non-goals honored

No changes to canonicalization, Ed25519 semantics, `AuthorizationV1` wire format, or the engine HMAC surface (#173 kept separate).

## Tests

New `verify.authorization.surface.test.ts` covers the #172 list:
- forged signature rejected in strict mode (and not marked verified)
- forged signature not reported as fully verified in default/permissive mode
- `signatureVerified === true` only when a cryptographic check actually ran and succeeded
- missing trusted key set does not silently produce a fully-verified result
- structure-only result is distinguishable from strict verification
- `signatureVerified` orthogonal to `ok` (valid-but-expired)

Core suite: **280 pass / 0 fail**.

## Validation

- `pnpm --filter @oxdeai/core typecheck` — passes
- core tests — 280 pass
- `pnpm -C packages/conformance validate` — 209 assertions pass
- `test:vectors:ts / auth / pep / delegation` — 11 / 12 / 9 / 12 pass
- `security-gate` — `Decision: ALLOW`
- `git diff --check` — clean

## API fingerprint note

The public API surface expands (three new exported types + the widened result), but `packages/core/API_FINGERPRINT` hashes only `dist/index.d.ts`, which is re-export-only and therefore **byte-identical** for this change (verified: a clean `upstream/main` build and this branch produce the same `index.d.ts` hash). I also noticed the committed fingerprint doesn't match a local clean build even without my change — consistent with a CRLF/LF or build-environment difference — so I deliberately left `API_FINGERPRINT` untouched rather than overwrite the canonical value from a Windows build. Happy to regenerate it in whatever way matches CI if you'd prefer it bumped.

### DIFF
--- a/packages/core/src/test/verify.authorization.surface.test.ts
+++ b/packages/core/src/test/verify.authorization.surface.test.ts
@@ -74,8 +74,10 @@ test("missing trusted key set does not silently produce a fully verified result"
   // verification occurred.
   const out = verifyAuthorization(makeAuth(), { now: 1010 });
   assert.equal(out.ok, true);
+  // Configured posture is permissive (best-effort default); the *absence* of a
+  // signature check is conveyed by signatureVerified/coverage, not the mode.
   assert.equal(out.signatureVerified, false);
-  assert.equal(out.verificationMode, "structure-only");
+  assert.equal(out.verificationMode, "permissive");
   assert.equal(out.verificationCoverage, "none");
 });
 
@@ -101,13 +103,15 @@ test("forged signature rejected in strict mode and not marked verified", () => {
   assert.ok(out.violations.some((v) => v.code === "AUTH_SIGNATURE_INVALID"));
 });
 
-test("explicit non-crypto (structure-only) result is distinguishable from strict verification", () => {
+test("permissive-without-crypto is distinguishable from strict verification", () => {
   const structural = verifyAuthorization(makeAuth(), { now: 1010 });
   const strict = verifyAuthorization(makeAuth(), { now: 1010, mode: "strict", trustedKeySets: KEYSET });
+  // Distinguished by posture (mode) AND by whether a signature was checked.
   assert.notEqual(structural.verificationMode, strict.verificationMode);
-  assert.equal(structural.verificationMode, "structure-only");
+  assert.equal(structural.verificationMode, "permissive");
   assert.equal(strict.verificationMode, "strict");
   assert.equal(structural.signatureVerified, false);
+  assert.equal(structural.verificationCoverage, "none");
   assert.equal(strict.signatureVerified, true);
 });
 
@@ -139,9 +143,10 @@ test("unknown kid engages no cryptographic check and is not marked verified", ()
   });
   assert.equal(out.ok, false);
   assert.equal(out.signatureVerified, false);
-  // No verify function ran (trust could not be resolved), so posture is
-  // structure-only rather than permissive.
-  assert.equal(out.verificationMode, "structure-only");
+  // Trust could not be resolved so no signature was checked; the posture is
+  // still the configured "permissive" — the missing check shows in
+  // signatureVerified/coverage, not the mode.
+  assert.equal(out.verificationMode, "permissive");
   assert.equal(out.verificationCoverage, "none");
   assert.ok(out.violations.some((v) => v.code === "AUTH_KID_UNKNOWN"));
 });
--- a/packages/core/src/test/verify.authorization.surface.test.ts
+++ b/packages/core/src/test/verify.authorization.surface.test.ts
@@ -1,0 +1,147 @@
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * verify.authorization.surface.test.ts
+ *
+ * Verification-surface descriptors (#172): a verification result must make it
+ * impossible to mistake a merely-structural check for a cryptographically
+ * verified authorization. These tests pin the contract that
+ * `signatureVerified` is true ONLY when a cryptographic check actually ran and
+ * succeeded, and that `verificationMode` / `verificationCoverage` describe the
+ * posture and covered surface honestly.
+ */
+import test from "node:test";
+import assert from "node:assert/strict";
+import { generateKeyPairSync } from "node:crypto";
+import { signAuthorizationEd25519, verifyAuthorization } from "../verification/index.js";
+import type { KeySet } from "../types/keyset.js";
+
+const KEYPAIR = generateKeyPairSync("ed25519", {
+  privateKeyEncoding: { format: "pem", type: "pkcs8" },
+  publicKeyEncoding: { format: "pem", type: "spki" },
+});
+
+const KEYSET: KeySet = {
+  issuer: "issuer-A",
+  version: "1",
+  keys: [{ kid: "2026-01", alg: "Ed25519", public_key: KEYPAIR.publicKey }],
+};
+
+function makeAuth(overrides?: { issued_at?: number; expiry?: number }) {
+  return signAuthorizationEd25519(
+    {
+      auth_id: "f".repeat(64),
+      issuer: "issuer-A",
+      audience: "rp-A",
+      intent_hash: "a".repeat(64),
+      state_hash: "b".repeat(64),
+      policy_id: "c".repeat(64),
+      decision: "ALLOW",
+      issued_at: overrides?.issued_at ?? 1000,
+      expiry: overrides?.expiry ?? 1060,
+      kid: "2026-01",
+    },
+    KEYPAIR.privateKey
+  );
+}
+
+test("signatureVerified === true only when cryptographic verification actually ran and succeeded", () => {
+  const out = verifyAuthorization(makeAuth(), {
+    now: 1010,
+    trustedKeySets: KEYSET,
+    requireSignatureVerification: true,
+  });
+  assert.equal(out.ok, true);
+  assert.equal(out.signatureVerified, true);
+  assert.equal(out.verificationMode, "permissive");
+  assert.equal(out.verificationCoverage, "authorization-v1-full");
+});
+
+test("strict mode with trustedKeySets reports strict posture and full coverage", () => {
+  const out = verifyAuthorization(makeAuth(), {
+    now: 1010,
+    mode: "strict",
+    trustedKeySets: KEYSET,
+  });
+  assert.equal(out.ok, true);
+  assert.equal(out.signatureVerified, true);
+  assert.equal(out.verificationMode, "strict");
+  assert.equal(out.verificationCoverage, "authorization-v1-full");
+});
+
+test("missing trusted key set does not silently produce a fully verified result", () => {
+  // Structurally valid, non-expired, but no trustedKeySets: the result is
+  // `ok: true` for backward compatibility, yet must NOT claim signature
+  // verification occurred.
+  const out = verifyAuthorization(makeAuth(), { now: 1010 });
+  assert.equal(out.ok, true);
+  assert.equal(out.signatureVerified, false);
+  assert.equal(out.verificationMode, "structure-only");
+  assert.equal(out.verificationCoverage, "none");
+});
+
+test("forged signature not reported as fully verified in default/permissive mode", () => {
+  // Tamper a signed field so the Ed25519 signature no longer matches, then
+  // verify in best-effort mode WITH a trusted key set (so a crypto check runs).
+  const forged = { ...makeAuth(), state_hash: "d".repeat(64) };
+  const out = verifyAuthorization(forged, { now: 1010, trustedKeySets: KEYSET });
+  assert.equal(out.ok, false);
+  assert.equal(out.signatureVerified, false);
+  assert.equal(out.verificationMode, "permissive");
+  assert.equal(out.verificationCoverage, "none");
+  assert.ok(out.violations.some((v) => v.code === "AUTH_SIGNATURE_INVALID"));
+});
+
+test("forged signature rejected in strict mode and not marked verified", () => {
+  const forged = { ...makeAuth(), audience: "rp-EVIL" };
+  const out = verifyAuthorization(forged, { now: 1010, mode: "strict", trustedKeySets: KEYSET });
+  assert.equal(out.ok, false);
+  assert.equal(out.signatureVerified, false);
+  assert.equal(out.verificationMode, "strict");
+  assert.equal(out.verificationCoverage, "none");
+  assert.ok(out.violations.some((v) => v.code === "AUTH_SIGNATURE_INVALID"));
+});
+
+test("explicit non-crypto (structure-only) result is distinguishable from strict verification", () => {
+  const structural = verifyAuthorization(makeAuth(), { now: 1010 });
+  const strict = verifyAuthorization(makeAuth(), { now: 1010, mode: "strict", trustedKeySets: KEYSET });
+  assert.notEqual(structural.verificationMode, strict.verificationMode);
+  assert.equal(structural.verificationMode, "structure-only");
+  assert.equal(strict.verificationMode, "strict");
+  assert.equal(structural.signatureVerified, false);
+  assert.equal(strict.signatureVerified, true);
+});
+
+test("signatureVerified is orthogonal to ok: validly-signed but expired is verified yet not ok", () => {
+  const out = verifyAuthorization(makeAuth({ issued_at: 1000, expiry: 1060 }), {
+    now: 2000, // past expiry
+    trustedKeySets: KEYSET,
+  });
+  assert.equal(out.ok, false);
+  assert.equal(out.signatureVerified, true);
+  assert.equal(out.verificationCoverage, "authorization-v1-full");
+  assert.ok(out.violations.some((v) => v.code === "AUTH_EXPIRED"));
+});
+
+test("strict mode without trustedKeySets reports strict posture, no signature verified", () => {
+  const out = verifyAuthorization(makeAuth(), { now: 1010, mode: "strict" });
+  assert.equal(out.ok, false);
+  assert.equal(out.signatureVerified, false);
+  assert.equal(out.verificationMode, "strict");
+  assert.equal(out.verificationCoverage, "none");
+  assert.ok(out.violations.some((v) => v.code === "TRUSTED_KEYSETS_REQUIRED"));
+});
+
+test("unknown kid engages no cryptographic check and is not marked verified", () => {
+  const out = verifyAuthorization({ ...makeAuth(), kid: "missing" }, {
+    now: 1010,
+    trustedKeySets: KEYSET,
+    requireSignatureVerification: true,
+  });
+  assert.equal(out.ok, false);
+  assert.equal(out.signatureVerified, false);
+  // No verify function ran (trust could not be resolved), so posture is
+  // structure-only rather than permissive.
+  assert.equal(out.verificationMode, "structure-only");
+  assert.equal(out.verificationCoverage, "none");
+  assert.ok(out.violations.some((v) => v.code === "AUTH_KID_UNKNOWN"));
+});
--- a/packages/core/src/test/verify.authorization.surface.test.ts
+++ b/packages/core/src/test/verify.authorization.surface.test.ts
@@ -81,25 +81,28 @@ test("missing trusted key set does not silently produce a fully verified result"
   assert.equal(out.verificationCoverage, "none");
 });
 
-test("forged signature not reported as fully verified in default/permissive mode", () => {
+test("forged signature not reported as verified, but coverage reflects the surface that was checked", () => {
   // Tamper a signed field so the Ed25519 signature no longer matches, then
   // verify in best-effort mode WITH a trusted key set (so a crypto check runs).
+  // The check DID execute against the full AuthorizationV1 payload, so coverage
+  // is "authorization-v1-full" even though authentication failed — the failure
+  // is conveyed by signatureVerified, keeping the two descriptors orthogonal.
   const forged = { ...makeAuth(), state_hash: "d".repeat(64) };
   const out = verifyAuthorization(forged, { now: 1010, trustedKeySets: KEYSET });
   assert.equal(out.ok, false);
   assert.equal(out.signatureVerified, false);
   assert.equal(out.verificationMode, "permissive");
-  assert.equal(out.verificationCoverage, "none");
+  assert.equal(out.verificationCoverage, "authorization-v1-full");
   assert.ok(out.violations.some((v) => v.code === "AUTH_SIGNATURE_INVALID"));
 });
 
-test("forged signature rejected in strict mode and not marked verified", () => {
+test("forged signature rejected in strict mode: not verified, full coverage (a check ran)", () => {
   const forged = { ...makeAuth(), audience: "rp-EVIL" };
   const out = verifyAuthorization(forged, { now: 1010, mode: "strict", trustedKeySets: KEYSET });
   assert.equal(out.ok, false);
   assert.equal(out.signatureVerified, false);
   assert.equal(out.verificationMode, "strict");
-  assert.equal(out.verificationCoverage, "none");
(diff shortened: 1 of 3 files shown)


AngeYobo:
This looks good overall. I agree with the direction and I think it addresses the integration footgun that motivated #172 while keeping the behavioral change separate.

A couple of comments before implementation:

1. Keep verificationMode independent from whether crypto happened

I agree with making the three facts orthogonal, but I’d avoid defining "structure-only" as a mode.

To me:

* verificationMode should answer how the verifier was configured (strict vs permissive).
* signatureVerified should answer whether a cryptographic signature was actually verified.
* verificationCoverage should answer what cryptographic surface was covered.

That way, a permissive verification that skipped signature verification because no trust anchor was available still reports:

verificationMode: "permissive"
signatureVerified: false
verificationCoverage: "none"

rather than inventing a third mode for what is really the absence of cryptographic verification.

Otherwise the meaning of verificationMode starts mixing execution posture with execution outcome.

⸻

2. I like defining engine-hmac-subset now

I agree with introducing verificationCoverage in #172 and reserving:

"engine-hmac-subset"

for #173.

That gives us one shared vocabulary across both verifiers instead of parallel status models.

I would just make it explicit in the docs that the AuthorizationV1 verifier never emits that value, so consumers don’t assume every verifier supports every coverage level.

⸻

Everything else looks right to me.

In particular I agree with:

* additive public fields rather than changing defaults;
* keeping strict-by-default as a separate fingerprint-reviewed issue;
* keeping signature validity orthogonal to overall authorization validity (signatureVerified: true, ok: false for an expired authorization);
* keeping #173 limited to clarifying the HMAC verification surface instead of expanding it.

The API surface is becoming much more explicit about what was verified, how it was verified, and what trust guarantees the caller actually has, which is exactly the direction I think the protocol should take.

Whatsonyourmind:
Thanks — both applied in 09c99f9.

1. **verificationMode is now configuration-only** (`"strict" | "permissive"`). Dropped `"structure-only"`; a permissive verification that skipped the signature check (no trust anchor) now reports `verificationMode: "permissive"`, `signatureVerified: false`, `verificationCoverage: "none"` — the absence of cryptography is conveyed by those two fields, not by the mode. The type doc now states the mapping explicitly (`best-effort` request → `permissive`) and that mode is independent of whether crypto ran.

2. **Documented that this verifier never emits `"engine-hmac-subset"`** — added to the `AuthorizationVerificationResult` doc, noting that value is reserved for `PolicyEngine.verifyAuthorization` (#173) and that consumers must not assume every verifier produces every coverage level.

No change to `signatureVerified` / `verificationCoverage` semantics or defaults. Re-validated: core tests 280/0, conformance 209, vectors (auth/pep/delegation/ts) pass, security-gate ALLOW, typecheck + `git diff --check` clean.

packages/core/src/test/verify.authorization.surface.test.ts
```diff
@@ -139,9 +143,10 @@ test("unknown kid engages no cryptographic check and is not marked verified", ()
...
-  // No verify function ran (trust could not be resolved), so posture is
-  // structure-only rather than permissive.
-  assert.equal(out.verificationMode, "structure-only");
+  // Trust could not be resolved so no signature was checked; the posture is
+  // still the configured "permissive" — the missing check shows in
+  // signatureVerified/coverage, not the mode.
+  assert.equal(out.verificationMode, "permissive");
   assert.equal(out.verificationCoverage, "none");
   assert.ok(out.violations.some((v) => v.code === "AUTH_KID_UNKNOWN"));
 });
```
AngeYobo, line 155: ### Review comment

**Please add a regression test pinning the strict-mode invariant.**

By inspection, strict mode can currently reach the HMAC branch with a non-empty `trustedKeySets` value but no `legacyHmacSecret`, and return successfully without any cryptographic verification.

A regression test should enforce:

```ts
verificationMode === "strict" && ok === true
```

always implies:

```ts
signatureVerified === true
verificationCoverage === "authorization-v1-full"
 ```

A concrete case worth testing is:

```ts
const auth = {
 ...makeAuth(),
  alg: "HMAC-SHA256" as const,
 };

const out = verifyAuthorization(auth, {
  now: 1010,
  mode: "strict",
  trustedKeySets: KEYSET,
});

assert.equal(out.ok, false);
assert.equal(out.signatureVerified, false);
assert.equal(out.verificationCoverage, "none");
assert.ok(
 out.violations.some((v) => v.code === "AUTH_TRUST_MISSING")
);
```

This protects the core invariant that strict verification must never succeed without a successful cryptographic check.

packages/core/src/verification/verifyAuthorization.ts
```diff
@@ -251,13 +251,13 @@ export function verifyAuthorization(
...
   //     actually runs and succeeds;
-  //   - `cryptoEngaged` records that a verify function was invoked at all
-  //     (pass or fail), which distinguishes "permissive" from "structure-only";
   //   - `verificationCoverage` records which surface that check covered.
-  // The default/permissive path never sets `signatureVerified`, so a caller
-  // that only inspects `ok` cannot be silently misled.
+  // `verificationMode` is derived purely from the requested posture (below),
+  // independent of whether cryptography ran — the absence of a signature check
+  // is conveyed by `signatureVerified: false` / `verificationCoverage: "none"`,
+  // not by the mode. The default/permissive path never sets `signatureVerified`,
```
AngeYobo: ### Review comment

> **This comment is no longer accurate.**
>
> A permissive call with a trusted key set and a valid signature does set `signatureVerified` to `true`, as covered by the first surface test.
>
> Suggested replacement:
>
> ```ts
> // A permissive result sets `signatureVerified` only when a cryptographic
> // check was available, executed, and succeeded.
> ```

packages/core/src/verification/verifyAuthorization.ts
```diff
@@ -376,7 +376,6 @@ export function verifyAuthorization(
           violations.push({ code: "AUTH_KEY_INACTIVE", message: "key is not active at verification time" });
         } else {
           // A cryptographic check is actually performed here (pass or fail).
-          cryptoEngaged = true;
           if (
             !verifyEd25519(SIGNING_DOMAINS.AUTH_V1, payload, sigValue, key.public_key) &&
```
AngeYobo, line 397: ### Review comment

> **Set coverage when the cryptographic verifier is engaged, not only after it succeeds.**
>
> The full AuthorizationV1 signing payload is evaluated in both the success and failure cases. `signatureVerified` should communicate the outcome; `verificationCoverage` should communicate the evaluated surface.
>
> Suggested structure:
>
> ```ts
> verificationCoverage = "authorization-v1-full";
>
> if (
>   !verifyEd25519(
>     SIGNING_DOMAINS.AUTH_V1,
>     payload,
>     sigValue,
>     key.public_key
>   ) &&
>   !verifyEd25519Raw(payload, sigValue, key.public_key)
> ) {
>   violations.push({
>     code: "AUTH_SIGNATURE_INVALID",
>     message: "signature verification failed"
>   });
> } else {
>   signatureVerified = true;
> }
> ```
>
> This produces the correct distinction:
>
> ```ts
> // Verifier ran and rejected the signature
> signatureVerified: false
> verificationCoverage: "authorization-v1-full"
>
> // No verifier ran
> signatureVerified: false
> verificationCoverage: "none"
> ```

packages/core/src/verification/verifyAuthorization.ts
```diff
@@ -392,7 +391,6 @@ export function verifyAuthorization(
       if (opts?.legacyHmacSecret) {
         // Legacy HMAC covers the full AuthorizationV1 signing payload (not the
         // narrower engine-HMAC subset), so successful coverage is "full".
-        cryptoEngaged = true;
         if (!verifyHmacDomain(SIGNING_DOMAINS.AUTH_V1, payload, sigValue, opts.legacyHmacSecret)) {
```
AngeYobo, line 412: ### Review comment

> The legacy HMAC branch should follow the same coverage semantics as Ed25519.
>
> Set:
>
> ```ts
> verificationCoverage = "authorization-v1-full";
> ```
>
> before calling `verifyHmacDomain`, then set only:
>
> ```ts
> signatureVerified = true;
> ```
>
> when the check succeeds.
>
> A failed HMAC check still evaluated the complete AuthorizationV1 signing payload and should not report coverage `"none"`.

packages/core/src/verification/verifyAuthorization.ts
AngeYobo: ### Review comment

> **Strict mode must implicitly require successful signature verification.**
>
> Currently, `requireSig` does not include `mode: "strict"`. This creates a fail-open path for legacy HMAC:
>
> * `mode: "strict"`;
> * non-empty `trustedKeySets`;
> * `alg: "HMAC-SHA256"`;
> * no `legacyHmacSecret`;
> * `requireSignatureVerification` unset.
>
> The strict precondition passes because `trustedKeySets` is non-empty, but the HMAC branch adds no violation because `requireSig` is false. The function can therefore return:
>
> ```ts
> ok: true
> verificationMode: "strict"
> signatureVerified: false
> verificationCoverage: "none"
> ```
>
> That contradicts the documented strict contract.
>
> Suggested correction:
>
> ```ts
> const requireSig =
>   opts?.mode === "strict" ||
>   (opts?.requireSignatureVerification ?? false);
> ```
>
> Please pair this with a regression test asserting that a strict successful result always has:
>
> ```ts
> signatureVerified === true
> verificationCoverage === "authorization-v1-full"
> ```

packages/core/src/test/verify.authorization.surface.test.ts
AngeYobo: ### Review comment

> **Coverage should describe what the cryptographic check evaluated, independently of whether the signature passed.**
>
> In this test, the Ed25519 verifier actually runs against the complete `AuthorizationV1` signing payload. The signature is rejected, so `signatureVerified` must remain `false`, but the evaluated surface was still the full authorization payload.
>
> I suggest changing the expectation to:
>
> ```ts
> assert.equal(out.signatureVerified, false);
> assert.equal(out.verificationCoverage, "authorization-v1-full");
> ```
>
> `"none"` should be reserved for cases where no cryptographic verifier ran at all, such as missing trust material, an unknown `kid`, or no HMAC secret.
>
> The same change should be applied to the strict forged-signature test below.

packages/core/src/verification/types.ts
```diff
@@ -3,26 +3,27 @@
...
+ * `verificationCoverage: "none"` — the absence of cryptography is conveyed by
+ * those fields, not by the mode.
  *
  * @public
  */
-export type VerificationMode = "strict" | "permissive" | "structure-only";
+export type VerificationMode = "strict" | "permissive";
 
 /**
  * Which authorization surface a signature check actually covered.
```
AngeYobo: ### Review comment

> **Please make the coverage semantics explicit for failed signature checks.**
>
> The three descriptors are intended to be orthogonal:
>
> * `signatureVerified`: whether authentication succeeded;
> * `verificationMode`: the configured verification posture;
> * `verificationCoverage`: which payload surface was evaluated cryptographically.
>
> Therefore, `"authorization-v1-full"` should apply whenever the full normative signing payload was submitted to the cryptographic verifier, regardless of whether the signature was accepted or rejected.
>
> Suggested wording:
>
> ```ts
> /**
>  * Which authorization surface was evaluated by a cryptographic verifier,
>  * independently of whether verification succeeded.
>  *
>  * - `"authorization-v1-full"` - the full normative AuthorizationV1 signing
>  *                               payload was evaluated cryptographically.
>  * - `"engine-hmac-subset"`    - only the engine-HMAC field subset was
>  *                               evaluated.
>  * - `"none"`                  - no cryptographic verifier ran.
>  *
>  * Inspect `signatureVerified` separately to determine whether the check
>  * succeeded.
>  */
> ```
>
> Without this clarification, `"none"` ambiguously represents both “not attempted” and “attempted but rejected.”

packages/core/src/verification/types.ts
AngeYobo: ### Review comment

> Since `VerificationMode` now represents configuration only, I suggest aligning this field comment with that contract:
>
> ```ts
> /** Configured verification posture. See {@link VerificationMode}. */
> ```
>
> “How verification was performed” can still be interpreted as runtime behavior, while the type documentation now correctly defines it as configuration-only.

AngeYobo (commented):
## Overall PR review summary

You can add this as the general review comment:

> The revised `verificationMode` model is correct: it now represents configuration only, while `signatureVerified` and `verificationCoverage` report runtime cryptographic facts.
>
> Before approval, I see three remaining items:
>
> 1. Correct the stale permissive-path comment.
> 2. Set coverage when a cryptographic verifier runs, regardless of success.
> 3. Ensure strict mode cannot return `ok: true` without successful full cryptographic verification.
>
> These changes preserve the intended orthogonality of the result descriptors and close a strict-mode fail-open edge case without changing canonicalization, wire format, or signing semantics.

AngeYobo:
Thanks, Luka.

The `verificationMode` update is correct and now cleanly represents the configured `strict` / `best-effort` posture independently from whether cryptographic verification actually ran.

After reviewing the updated implementation, I still see three items to address before approval.

### 1. Stale implementation comment

This comment is no longer accurate:

```ts
// The default/permissive path never sets `signatureVerified`
```

A permissive call with trusted key material and a valid signature does set `signatureVerified` to `true`, as confirmed by the first surface test.

Suggested replacement:

```ts
// A permissive result sets `signatureVerified` only when a cryptographic
// check was available, executed, and succeeded.
```

### 2. Coverage is still coupled to successful authentication

The forged-signature tests currently expect:

```ts
signatureVerified: false
verificationCoverage: "none"
```

However, in those cases the cryptographic verifier did execute against the complete `AuthorizationV1` signing payload. The result should therefore be:

```ts
signatureVerified: false
verificationCoverage: "authorization-v1-full"
```

`signatureVerified` should report whether authentication succeeded, while `verificationCoverage` should report which surface was evaluated cryptographically, independently of the outcome.

Please set the coverage when the cryptographic check is engaged, before evaluating whether it succeeds. Reserve `"none"` for cases where no cryptographic verifier ran, such as missing trust material, an unknown `kid`, or an absent HMAC secret.

Otherwise, coverage cannot distinguish between:

* cryptographic verification not attempted; and
* cryptographic verification attempted but rejected.

That would leave the descriptors partially coupled rather than fully orthogonal.

### 3. Strict HMAC verification appears able to succeed without cryptography

`requireSig` currently ignores `mode: "strict"`:

```ts
const requireSig = opts?.requireSignatureVerification ?? false;
```

Consider this path:

* `mode: "strict"`;
* a non-empty `trustedKeySets`;
* an authorization using `HMAC-SHA256`;
* no `legacyHmacSecret`;
* `requireSignatureVerification` unset.

The initial strict-mode check passes because `trustedKeySets` is non-empty. The HMAC branch then adds no violation because `requireSig` remains `false`.

By inspection, that can produce:

```ts
ok: true
verificationMode: "strict"
signatureVerified: false
verificationCoverage: "none"
```

That contradicts the documented strict-mode contract.

Suggested correction:

```ts
const requireSig =
  opts?.mode === "strict" ||
  (opts?.requireSignatureVerification ?? false);
```

Please also add a regression test pinning the invariant that:

```ts
verificationMode === "strict" && ok === true
```

always implies:

```ts
signatureVerified === true
verificationCoverage === "authorization-v1-full"
```

Once these three points are addressed, the PR should be ready for approval.

Whatsonyourmind:
Thanks — all three addressed in `22f1c38`.

**1. Stale comment.** Replaced; the descriptor block now states that a permissive result sets `signatureVerified` only when a cryptographic check was available, executed, and succeeded.

**2. Coverage decoupled from outcome.** This is the better model — I set `verificationCoverage = "authorization-v1-full"` the moment the cryptographic check is *engaged* (in both the Ed25519 and legacy-HMAC branches), before evaluating success. So a forged signature now reports `verificationCoverage: "authorization-v1-full"` with `signatureVerified: false`, and `"none"` is reserved strictly for the cases where no verifier ran at all: missing trust material, unknown/inactive `kid`, or absent HMAC secret. The two descriptors are now fully orthogonal — `signatureVerified` = did authentication succeed, `verificationCoverage` = which surface was cryptographically evaluated. Updated the two forged-signature surface tests and the shared type docs to match.

**3. Strict implies signature verification.** Applied your `requireSig` correction:

```ts
const requireSig =
  opts?.mode === "strict" ||
  (opts?.requireSignatureVerification ?? false);
```

Added a regression test pinning the invariant directly (`verificationMode === "strict" && ok === true` ⟹ `signatureVerified === true && verificationCoverage === "authorization-v1-full"`), plus a negative test for the exact gap you identified: `mode: "strict"` + non-empty `trustedKeySets` + `HMAC-SHA256` + no `legacyHmacSecret` now yields `AUTH_TRUST_MISSING` (`ok: false`) rather than an unauthenticated `ok: true`.

No change to signing payloads, canonicalization, or the default (permissive) accept/reject behavior. Gates green: `@oxdeai/core` typecheck + 282 tests, conformance 209 assertions, all four vector suites (ts/auth/pep/delegation), security-gate ALLOW, `git diff --check` clean.

AngeYobo:
Thanks, Luka. I reviewed 22f1c38, and all three blocking points are now correctly addressed.

* The stale permissive-path comment is fixed.
* Coverage is now recorded when the cryptographic verifier is engaged, independently of authentication success, for both Ed25519 and legacy HMAC.
* Strict mode now makes signature verification mandatory, and the new tests pin both the positive `strict + ok` invariant and the exact missing-HMAC-secret bypass case.

The resulting contract is now cleanly orthogonal:

* `verificationMode` describes configured posture;
* `signatureVerified` describes authentication success;
* `verificationCoverage` describes the surface evaluated cryptographically.

No remaining blockers from my review. Approved.

Two optional cleanup nits, not required for merge:

* the success branches can remove the second assignment of `verificationCoverage = "authorization-v1-full"` because it is already set before the check;
* the shared field comment could say “Configured verification posture” rather than “How verification was performed.”

Neither affects correctness or approval.

Whatsonyourmind:
Thanks for the review and the approval — appreciate the careful pass on the orthogonality contract.

On the two nits:

- **Second coverage assignment** — already gone in `22f1c38`: moving `verificationCoverage = "authorization-v1-full"` ahead of the check meant the success branches now only set `signatureVerified = true`, so there's no duplicate assignment left to remove in either the Ed25519 or HMAC path.
- **Field comment wording** — applied in `8d804b2`: the shared `verificationMode` field now reads "Configured verification posture" to match the type-level doc.

Doc-only follow-up, no behavior change; gates still green (typecheck + 282 core tests, whitespace clean). Ready to merge whenever suits you.

AngeYobo:
Thanks, Luka. Confirmed.

You’re right regarding the coverage assignment. My nit came from reading the old/new lines in the GitHub diff rather than the final file state. 22f1c38 already leaves only signatureVerified = true in the success branches.

The wording update in 8d804b2 also aligns the field-level documentation with the configuration-only semantics of verificationMode.

No further changes requested. The PR is approved and ready to merge.

Issue #198, opened 2026-07-21
P1: Remediate newly disclosed brace-expansion and js-yaml dependency advisories


AngeYobo:
## Summary

The repository security gate is now failing because two newly disclosed high-severity advisories affect transitive development dependencies already present in the current lockfile:

```text
brace-expansion
Advisory ID: 1123898
Severity: high
Path:
packages/core
  > @microsoft/api-extractor@7.58.7
  > minimatch@10.2.5
  > brace-expansion@5.0.6
```

```text
js-yaml
Advisory ID: 1123911
Severity: high
Path:
repository root
  > puppeteer@24.43.0
  > cosmiconfig@9.0.1
  > js-yaml@4.2.0
```

The security gate currently reports:

```text
== Security Gate ==

Findings: 3
Blocking: 2
Matched exceptions: 0
Expired exceptions: 0
Warnings: 1

Decision: DENY
Reason: policy denies this severity
```

These dependencies were already present in the lockfile when the gate last passed. The failure is caused by newly published advisory data rather than a dependency change introduced by the currently open PR.

## Priority

**P1**

The affected dependencies appear to be development and tooling dependencies rather than runtime protocol dependencies. However, the advisories are high severity and currently block all pull requests through the repository security gate.

## Objective

Restore a passing security gate by remediating both advisories through supported dependency upgrades or narrowly scoped package-manager overrides.

The preferred outcome is to remove the vulnerable versions from the resolved dependency graph.

Security-policy exceptions should be used only if a safe upgrade is not currently possible and the actual exploitability has been assessed and documented.

## Required investigation

Before changing the dependency graph, determine:

1. Whether newer direct versions of the parent packages resolve the advisories:

   * `@microsoft/api-extractor`
   * `puppeteer`
   * `cosmiconfig`, if directly controllable

2. Whether the affected transitive packages have patched versions compatible with the current parents:

   * `brace-expansion`
   * `js-yaml`

3. Whether a `pnpm.overrides` entry can safely force patched versions without breaking:

   * API extraction
   * build tooling
   * documentation generation
   * Puppeteer-based scripts or tests

4. Whether either advisory is unreachable in the repository’s actual usage.

5. Whether the advisories affect:

   * production/runtime dependencies;
   * published package contents;
   * development-only execution;
   * CI-only tooling.

## Preferred remediation order

Use the first viable option:

### Option A:  Upgrade the direct parent dependency

[...]

Issue #197, opened 2026-07-21
P1: Authenticate evaluator input provenance (state source + self-declared intent fields)


AngeYobo:
## Summary

OxDeAI cryptographically protects the **authorization verdict** through signatures, bindings, replay protection, and deterministic verification. It does not yet authenticate every **premise** consumed by the deterministic evaluator.

A signed `state_hash` and `policy_id` establish consistency at the time of evaluation. They do not establish that the underlying state was authoritative, current, or independently obtained. If the same compromised runtime supplies both the intent and the “live” state later re-hashed by the guard, successful verification proves only that the supplied values are mutually consistent.

This issue addresses a remaining trust-boundary gap with two manifestations and one root cause.

### 1. State-source provenance

This is currently documented as RT-TRUST-1.

In Profile C, the guard recomputes the hash of the live state and compares it with the committed `state_hash`. This proves that the evaluated object matches the commitment. It does not prove that the object originated from an authoritative and current state source.

A compromised orchestrator can therefore:

1. construct an intent using false or stale state;
2. submit the same state object to the guard;
3. satisfy the `state_hash` comparison.

The binding is valid, but the bound premise is not authenticated.

### 2. Self-declared intent fields

`agent_id` and several structural fields, including `depth`, `tool_call`, and `tool`, are currently supplied by the proposer.

`intent.signature` is not load-bearing in the current engine and PEP reference paths, so these values are not independently authenticated before policy evaluation. A compromised runtime may therefore influence not only the proposed action, but also which per-agent or structural policy premises are applied.

This behavior is reproduced in:

```text
scripts/security/repro-policy-boundary-attacks.ts
````

Relevant cases:

```text
G / H / L / M
```

### Root cause

The agent should be able to influence the **intent proposal**, but not the trusted premises against which that proposal is evaluated.

Today, the same runtime may influence both.

This issue tracks reducing that input surface so that every evaluator premise is either:

* obtained from a trusted authority;
* cryptographically authenticated;
* derived from trusted enforcement context; or
* explicitly treated as untrusted proposal data.

## Why this is not already covered

* The trusted-time work in #191- #195 closes one instance of the same class: the timestamp premise. It replaces proposer-controlled time with trusted `evaluation_time`. This issue addresses the remaining premises, particularly state, identity, and structural context.
* #166 provides policy and state-modeling guidance. This issue is the enforcement and protocol counterpart: it changes what the reference boundary verifies rather than only documenting deployment expectations.
[...]

Issue #195, opened 2026-07-21
P2: Stop using intent.timestamp as verifier time in the SDK client


AngeYobo:
<html>
<body>
<html><head></head><body><h2>Summary</h2><p>Update the SDK so it no longer passes attacker-controlled <code inline="">intent.timestamp</code> as the verifier's notion of "current time" when verifying AuthorizationV1 artifacts.</p><p>The audit identified SDK logic equivalent to:</p><pre><code class="language-ts">verifyAuthorization(
  authorization,
  trustedKeys,
  intent.timestamp
);
</code></pre><p>or:</p><pre><code class="language-ts">createVerifier(...).verify(
  authorization,
  intent.timestamp
);
</code></pre><p>This makes the SDK verification path depend on an untrusted timestamp supplied by the agent rather than a trusted verifier clock.</p><p>The SDK must instead require an explicit trusted verification time or obtain one from the trusted execution boundary.</p><p>Refs #171.</p><hr><h1>Priority</h1><p><strong>P2</strong></p><p>This is primarily an SDK correctness and API-hardening issue.</p><p>It does <strong>not</strong> affect the core verifier itself, which must independently enforce trusted-time semantics. Instead, it prevents SDK consumers from accidentally supplying attacker-controlled verifier time.</p><hr><h1>Problem statement</h1><p>Authorization verification has two independent clock domains:</p>
Clock | Trust
-- | --
intent.timestamp | Untrusted
verifier time | Trusted

<p>The SDK currently conflates them.</p><p>As a consequence, SDK consumers can unknowingly write:</p><pre><code class="language-ts">verifyAuthorization(
  auth,
  keySet,
  intent.timestamp
);
</code></pre><p>which effectively allows the intent to define the verifier's current time.</p><p>Even after trusted-time issuance lands, this remains an incorrect verification model.</p><hr><h1>Objective</h1><p>The SDK must make trusted verifier time explicit.</p><p>After this issue, SDK consumers should either:</p><pre><code class="language-ts">verifyAuthorization(
  auth,
  keySet,
  trustedNow
);
</code></pre><p>or:</p><pre><code class="language-ts">verifyAuthorization(
  auth,
  keySet,
  {
    verificationTime: trustedNow
  }
);
[...]

PR #202, opened 2026-07-25
feat(policy): clarify PolicyEngine.verifyAuthorization as limited-scope engine-HMAC verification (#173)

### LINKED ISSUE #173: P1: Clarify or harden engine.verifyAuthorization HMAC verification surface

Whatsonyourmind:
## Summary

Resolves #173 via the **clarified limited-scope path** you confirmed. `PolicyEngine.verifyAuthorization` authenticates only the engine-HMAC field subset (`intent_hash`, `policy_version`, `state_snapshot_hash`, `decision`, `expires_at`), so an integrator can mistake it for full authorization verification. This PR makes that boundary explicit rather than expanding the HMAC (which would change canonicalization/vectors and duplicate the strict verifier).

## Contract (as agreed)

Reusing the #172 vocabulary, the result now **always** carries the verification-surface descriptors, **required** on a dedicated result type rather than merely inherited as optional fields from the shared `VerificationResult`:

- New `EngineAuthorizationVerificationResult` = the existing `{ valid, reason? }` **plus required** `signatureVerified` / `verificationMode` / `verificationCoverage`.
- A successful engine-HMAC check reports:
  ```ts
  signatureVerified: true
  verificationMode: "permissive"
  verificationCoverage: "engine-hmac-subset"
  ```
  It never emits `"authorization-v1-full"`.
- A failed or unperformed HMAC never claims successful verification: `signatureVerified: false`. `verificationCoverage` is `"engine-hmac-subset"` once the check is engaged (pass **or** fail), and `"none"` when a structural check short-circuits before any HMAC runs.
- Docs state clearly that `ok`/`valid: true` authenticates only the engine-HMAC subset and is **not** equivalent to full `AuthorizationV1` verification; enforcement / relying-party integrations are directed to the strict standalone `verifyAuthorization`.

## Tests — framed as non-coverage demonstrations

`verify.engine-authorization.surface.test.ts` pins the distinction between engine-HMAC subset authentication and full AuthorizationV1 enforcement verification, explicitly as limitations rather than desirable acceptance:

- mutations to `audience` / `capability` / `state_hash` remain `valid` under the engine helper (outside the authenticated subset);
- the **strict** verifier rejects the same field-class mutations (contrast test);
- an in-subset mutation (`state_snapshot_hash`) invalidates the engine result with `signatureVerified: false`, `verificationCoverage: "engine-hmac-subset"`;
- a tampered engine HMAC reports `signatureVerified: false`;
- a structural short-circuit (expired) reports `verificationCoverage: "none"`.

## Boundary

No canonicalization, conformance-vector, signing-payload, Ed25519, trusted-time, or default-behavior changes. The `valid` / `reason` outputs are unchanged — the descriptors are strictly additive to the returned object.

## Validation

- `pnpm --filter @oxdeai/core typecheck` — clean
- `pnpm --filter @oxdeai/core test` — **288 pass** (6 new)
- `pnpm -C packages/conformance validate` — 209 assertions pass
- `pnpm test:vectors:{ts,auth,pep,delegation}` — 11 / 12 / 9 / 12 pass
- `git diff --check` — clean

### `security:gate`
Currently **DENY**, but from two pre-existing repo-wide dependency advisories unrelated to this PR's source-only diff (index.ts, PolicyEngine.ts, one new test — no `package.json`/lockfile changes):

- `high | js-yaml | id=1124281 | . > puppeteer > cosmiconfig > js-yaml@5.2.1`
- `high | brace-expansion | id=1124334 | packages/core > @microsoft/api-extractor > minimatch > brace-expansion@5.0.7`

Same class you cleared for #196 (#198). I've kept dependency remediation out of this feature PR per your preference — happy to rebase once the advisory gate is resolved.

### `API_FINGERPRINT`
This adds one public export (`EngineAuthorizationVerificationResult`) and widens the engine method's return type, so `packages/core/dist/index.d.ts` — and therefore the fingerprint — legitimately changes. My local (Windows) build hashes differ from the committed canonical value even for the *unchanged* upstream surface (`0abe9d4b` locally vs. committed `832c8fab`), so I did **not** overwrite `packages/core/API_FINGERPRINT` from a non-canonical build. Please regenerate it authoritatively on CI/Linux (`pnpm -C packages/core api:fingerprint`); for reference my local build moves it `0abe9d4b → 36144da2` with this change, confirming the surface delta.


AngeYobo:
Thanks, Luka. The limited-scope implementation and result contract look aligned with the agreed direction for #173.

I agree with keeping dependency remediation outside this feature PR. The two blocking advisories are repository-wide and unrelated to the `PolicyEngine.verifyAuthorization` source changes, but the security policy is fail-closed, so we should not merge #202 while the gate reports `DENY`.

Please keep #202 unchanged for now. I’ll handle the blocking dependency advisories in a separate remediation PR:

* `js-yaml` through `puppeteer > cosmiconfig`
* `brace-expansion` through `@microsoft/api-extractor > minimatch`

Once that remediation is merged, please rebase #202 onto the updated `main` and rerun:
core typecheck
core tests
conformance
all vector suites
security:gate
git diff --check

On the API fingerprint: agreed not to commit a fingerprint generated from a locally inconsistent Windows baseline. Because this PR legitimately changes the public API, the fingerprint must be regenerated from the authoritative Linux/CI build after the rebase. We should not use the local `0abe9d4b → 36144da2` values as the canonical repository value.

No need to mix dependency or fingerprint-environment fixes into #202. Once the security remediation lands and CI produces the authoritative fingerprint, we can complete the review and merge.

Whatsonyourmind:
Sounds good, @AngeYobo — I'll leave #202 exactly as-is and won't touch the dependencies or the fingerprint. Once your remediation PR for the `js-yaml` (puppeteer → cosmiconfig) and `brace-expansion` (@microsoft/api-extractor → minimatch) advisories lands, I'll rebase #202 onto the updated `main` and rerun the full gate — core typecheck, core tests, conformance, all four vector suites, `security:gate`, and `git diff --check` — then confirm here, leaving the authoritative `API_FINGERPRINT` for the Linux/CI build to regenerate. Ping me whenever it's merged.

Whatsonyourmind:
Rebased onto `main` (now including #204) — clean, no conflicts. Re-ran the full gate locally, all green:

- core typecheck — clean
- core tests — pass
- conformance — 209 assertions pass
- vectors — ts 11 / auth 12 / pep 9 / delegation 12
- `security:gate` — **Decision: ALLOW** (0 blocking; #204 cleared the `js-yaml` + `brace-expansion` advisories)
- `git diff --check` — clean

Head is now `7d808b7`; the commit is unchanged (author + DCO sign-off intact), and the diff carries no dependency or fingerprint edits. As agreed, I've left `API_FINGERPRINT` for the authoritative Linux/CI build to regenerate. Over to CI and your review — thanks for handling the advisory remediation.

AngeYobo:
Thanks, Luka. The rebase and validation results look good.

Before squash merge, there are two final items:

API_FINGERPRINT still needs to be regenerated and committed from an authoritative Linux environment. #202 publicly exports EngineAuthorizationVerificationResult and changes the return type of PolicyEngine.verifyAuthorization, so the API declaration surface legitimately changed.

In verify.engine-authorization.surface.test.ts, the contrast test is described as exercising the strict verifier, but currently calls verifyAuthorization without mode: "strict". Please update it to:
const strict = verifyAuthorization(forged, {
   now: 1010,
   mode: "strict",
   trustedKeySets: KEYSET
});

assert.equal(strict.verificationMode, "strict");

Once the fingerprint is committed, that test accurately pins strict mode, and CI is fully green, the PR is ready for squash merge.

Whatsonyourmind:
Done on the test — folded into the branch (now `aa94228`, still a single DCO-signed commit). The CONTRAST test now calls:

```ts
const strict = verifyAuthorization(forged, {
  now: 1010,
  mode: "strict",
  trustedKeySets: KEYSET,
});
...
assert.equal(strict.verificationMode, "strict");
```

so it accurately pins strict mode. Full core suite still green (288 pass), typecheck clean, `git diff --check` clean.

On the fingerprint — this is the one item I can't do faithfully from here, and I'd rather flag it than write a wrong value into the guard. My environment is Windows and its build is non-canonical: even on the *unchanged* upstream surface it hashes `0abe9d4b…` locally versus the committed `832c8fab…`, so any value I generate here would be inconsistent with the authoritative Linux one (my WSL has no Node toolchain either). Since #202 legitimately changes the public API surface (the `EngineAuthorizationVerificationResult` export + the `verifyAuthorization` return type), the new value needs to come from an authoritative Linux/CI build.

Could you regenerate and commit it at squash time (`pnpm -C packages/core api:fingerprint`), or share the CI-produced hash and I'll commit exactly that? Either works — I just don't want a Windows-baseline value in the file.

AngeYobo:
Thanks, Luka ! #173 is merged and closed, with the API fingerprint updated
and all gates green.

I reviewed the remaining open issues and suggest we continue with the
trusted-time series in this order:

1. #191 - drive replay-window eviction from trusted `evaluation_time`
2. #192 - drive velocity windows from trusted `evaluation_time`
3. #193 - mint `AuthorizationV1` timestamps from `evaluation_time`
4. #194 - execute the trusted-time conformance vectors
5. #195 - remove `intent.timestamp` as verifier time from the SDK client

This keeps the sequence layered: core stateful enforcement first,
authorization issuance second, conformance once semantics are stable, and
SDK migration last.

After that, we should take #197, probably splitting state provenance from
identity/structural provenance before implementation.

Please pick up #191 next, keeping it isolated from velocity, issuance,
conformance, and SDK changes.

Issue #192, opened 2026-07-21
P1: Drive velocity windows from `evaluation_time`


AngeYobo:
## Summary

Change velocity-window accounting and reset logic so they are driven by trusted `evaluation_time`, not attacker-controlled `intent.timestamp`.

The current velocity module derives its effective clock from:

```ts
const now = intent.timestamp;
```

It then uses that value to determine whether a velocity window has expired and whether the action counter should reset.

Because `intent.timestamp` is supplied by the agent, it must not control rate-limit window progression. The trusted-time freshness gate now bounds how far an intent timestamp may diverge from `evaluation_time`, but this is only a compensating control. It does not satisfy the stronger invariant that security-sensitive velocity state transitions are controlled by trusted time.

This issue implements that direct invariant.

Refs #171.

---

## Priority

**P1**

The freshness gate blocks the previously demonstrated large future-dating bypass. However, velocity-window state still depends directly on untrusted intent data and remains inconsistent with the Trusted-Time Authorization Core Profile.

---

## Current behavior

The velocity module currently follows this pattern:

```ts
const now = intent.timestamp;
const windowExpired =
  now >= currentWindowStart + configuredWindowSeconds;
```

When the module considers the window expired, it resets or replaces velocity state using the intent-supplied timestamp.

This permits an agent-controlled value to influence:

* whether the current velocity window remains active;
* whether the action count resets;
* the start time recorded for a replacement window;
* whether an action is allowed or denied.

The upstream freshness gate limits this manipulation to the configured skew and age bounds, but it does not remove the attacker-controlled timestamp from the velocity algorithm.

---

## Required invariant

All velocity-window calculations must use trusted `evaluation_time`.

Conceptually:

```text
window_end =
    trusted_window_start
  + configured_window_seconds
```

and:

```text
window_expired =
    evaluation_time >= window_end
```

The following must not determine velocity-window reset or progression:

* `intent.timestamp`;
* authorization `issued_at`;
* authorization `expiry`;
* timestamps supplied inside an intent or authorization;
* ambient `Date.now()` inside deterministic policy evaluation.

The intent timestamp remains subject to the freshness gate but must not function as the velocity clock.

---

## Security objective

A caller must not be able to:

* reset a velocity counter;
* advance the current velocity window;
* create a replacement window;
* obtain additional quota;

by modifying `intent.timestamp`.

For an active window beginning at trusted time `T₀`, with duration `W`:

```text
evaluation_time < T₀ + W
```

[...]

Issue #191, opened 2026-07-21
P1: Drive replay-window eviction from trusted evaluation_time


AngeYobo:
## Summary

Change replay-window retention and nonce eviction so they are driven by trusted `evaluation_time`, not by attacker-controlled `intent.timestamp`.

The current replay module computes its effective time from:

```ts
const now = intent.timestamp;
```

and then derives the active replay window from that value.

Because `intent.timestamp` is supplied by the agent, it must not directly control nonce retention or eviction. The trusted-time freshness gate now limits how far an intent timestamp may diverge from `evaluation_time`, but this is only a compensating bound. It does not satisfy the stronger invariant that replay state transitions are controlled by trusted time.

This issue implements that direct invariant.

Refs #171.

---

## Priority

**P1**

The known large future-dating bypass is currently blocked by the trusted-time freshness gate. However, replay retention still depends on untrusted intent data and remains inconsistent with the trusted-time specification.

---

## Current behavior

The replay module currently follows this shape:

```ts
const now = intent.timestamp;
const windowStart = now - cfg.window_seconds;
```

Retained nonce entries older than `windowStart` are excluded or evicted.

This means the intent controls the time coordinate used to determine whether an earlier nonce remains protected.

The upstream freshness gate enforces:

```text
intent.timestamp <= evaluation_time + maxClockSkewSeconds
intent.timestamp >= evaluation_time - maxIntentAgeSeconds
```

That limits exploitation but does not eliminate the trust-boundary violation.

---

## Required invariant

All replay-window calculations must use trusted `evaluation_time`:

```text
window_start = evaluation_time - effective_retention_window
```

The following must not determine replay eviction:

* `intent.timestamp`;
* authorization `issued_at`;
* authorization `expiry`;
* any timestamp supplied inside the artifact being evaluated;
* ambient `Date.now()` inside deterministic policy evaluation.

The intent timestamp may still be evaluated by the freshness gate, but it must not serve as the replay-store clock.

---

## Security objective

A caller must not be able to shorten nonce retention, evict a retained nonce, or alter the replay window by changing `intent.timestamp`.

For a nonce retained at trusted time `T₀`:

```text
same nonce evaluated at T₁
```

must be rejected whenever:

```text
T₁ - T₀ <= effective_retention_window
```

regardless of the timestamp declared by the second intent, provided the request reaches replay evaluation.

---

## Trusted time source

Use the same explicit `evaluation_time` supplied to deterministic policy evaluation.

[...]

PR #205, opened 2026-07-28
feat(policy): drive replay-window eviction from trusted evaluation time (#191)

Whatsonyourmind:
Implements #191 on the basis you confirmed: `evaluationTime` threaded through the existing decision context, replay eviction driven exclusively by the trusted clock, other modules receiving the value but behaviorally unchanged, and the self-healing no-migration path for persisted state.

**1. Problem.** `ReplayModule` derived both its pruning window and its nonce stamps from `intent.timestamp`, which is attacker-controlled. Backdating an intent moved `windowStart` before the caller's own recorded nonce and evicted it, turning a replay into an ALLOW; postdating parked a nonce beyond the retention horizon.

**2. Semantics.** Exactly as specified:

```
windowStart = evaluationTime - window_seconds
ts          = evaluationTime            // newly recorded nonces
```

No `intent.timestamp`, no `issued_at`, no `expiry`, no ambient `Date.now()`.

**3. The seam.** `evaluationTime` is threaded, never sampled. `evaluatePure` already validates it with `assertProtocolSeconds` before any module runs, and now passes it into `DecisionInput`. `runDecisionModules` builds one `PolicyEvaluationContext` per evaluation and hands the *same instance* to every module, so the clock cannot drift between modules within one evaluation. The phase stays pure and replayable because the value arrives as input.

**4. On the API point you asked me to verify.** I checked it rather than assuming, and the answer is in two parts.

Adding a third parameter does *not* force external implementations to widen: TypeScript treats a function with fewer parameters as assignable to a function type with more. The proof is in the build — the eight unchanged modules are still registered as bare two-parameter functions, and `decision.test.ts`'s stubs use zero-parameter `evaluate: () => …`. Both compile untouched. The only typecheck error the change produced anywhere was a test helper constructing `DecisionInput` literally.

But your extensibility concern stands on its own, so I took the context object anyway:

```ts
export interface PolicyEvaluationContext {
  evaluationTime: number;
}
```

`#192`–`#195` then add *fields* rather than widening the signature again. Adopting it now also means external modules that do opt into the third parameter type it once and never touch it again.

**5. Scope.** One module changed. `git diff --name-only upstream/main` touches exactly one file under `policy/modules/` — `ReplayModule.ts`. The other eight receive the context and are behaviorally unchanged, per your instruction to keep #191 isolated.

**6. Persisted-state compatibility.** Self-healing, no migration. No entry is rewritten, nothing is globally invalidated, and there is no schema or version bump. The contract is documented on `ReplayModule` itself and pinned by tests, in your words:

- legacy replay entries may have been stamped from `intent.timestamp`;
- they are interpreted as existing persisted timestamps, without conversion;
- all new entries are stamped from `evaluationTime`;
- pruning is driven exclusively by `evaluationTime`;
- no persisted-state schema migration is required;
- after one replay retention window, all surviving entries use trusted-time stamps.

Safe because the freshness gate already bounds the deviation between the two clocks, so the transitional discrepancy is bounded by the configured tolerance and ages out within one window.

**7. Tests.** New `replay.trusted-time.test.ts`, 8 cases, pinning the three properties separately because they regress independently: the window follows the trusted clock (both directions — a stale entry is pruned even when `intent.timestamp` would retain it, and a fresh one survives even when `intent.timestamp` would evict it); new nonces carry `evaluationTime`; and the compatibility path (legacy `ts` preserved verbatim, a legacy nonce still inside the window blocks its own replay, and after one window only trusted-stamped entries survive). One test reproduces the backdating attack directly and asserts `REPLAY_NONCE`.

Every fixture keeps `intent.timestamp` numerically distinct from `evaluationTime` — with both set to the same value these tests would pass against the old code and prove nothing. `baseInput()` in `decision.test.ts` got the same treatment (`evaluationTime` deliberately offset from the intent clock).

**8. Not touched: `ReplayEngine`.** Worth stating since the names collide — `replay/ReplayEngine.ts` is deterministic audit re-execution, not nonce replay-protection. It already threads `evaluationTime` into `evaluatePure` and needed no change.

**9. Validation.** Full suite, exact counts:

| gate | result |
|---|---|
| core typecheck | clean |
| core tests | **296 pass / 0 fail** (288 before, +8 new) |
| workspace typecheck | clean, all packages |
| conformance | **209 assertions passed** |
| vectors ts / auth / pep / delegation | **11 / 12 / 9 / 12**, no FAIL |
| `security:gate` | Findings 1, **Blocking 0**, Decision **ALLOW** |
| `repro-policy-boundary-attacks` | **byte-identical to `upstream/main`** |
| lint | clean |
| `git diff --check` | clean |

For the boundary repro I captured the output on `upstream/main` and on this branch and diffed them, rather than just eyeballing that it still runs — the two are identical, so the change moves no boundary.

**10. One local caveat, not a CI signal.** `pnpm test` at the workspace root fails on my Windows box inside `tests/`, where the script is `node --test $(find dist … | sort)` — the command substitution goes through `cmd.exe` and dies with `"sort)" non è riconosciuto`. Running the same tests directly gives **27 pass / 0 fail**. I did not touch `tests/package.json`; flagging it only so a Windows contributor isn't puzzled by it, and I'd rather report it than quietly omit a red.

**11. API_FINGERPRINT.** Left for CI. This is a legitimate delta: `PolicyEvaluationContext` is new `@public` surface and `PolicyModule.evaluate` gains its third parameter, both of which appear in `core.api.md`. I did not commit a value — as on #202, my Windows build is non-canonical, so the authoritative hash needs regenerating on Linux.

**12. Follow-ups deliberately excluded.** Nothing from #192–#195 is anticipated here: no velocity change, no AuthV1 timestamp work, no conformance-vector edits, no SDK surface. The context object is the only shared groundwork this PR lays.

### DIFF
--- a/packages/core/src/policy/PolicyEngine.ts
+++ b/packages/core/src/policy/PolicyEngine.ts
@@ -479,7 +479,7 @@ export class PolicyEngine {
 
       // ── Decision phase ──────────────────────────────────────────────────
       // (intent + state + module policies) → ALLOW | DENY + nextState
-      const decisionResult = runDecisionModules({ intent, state, mode }, modules);
+      const decisionResult = runDecisionModules({ intent, state, evaluationTime, mode }, modules);
 
       if (decisionResult.decision === "DENY") {
         this.emitAudit({
--- a/packages/core/src/policy/decision/runDecisionModules.ts
+++ b/packages/core/src/policy/decision/runDecisionModules.ts
@@ -1,22 +1,27 @@
 // SPDX-License-Identifier: Apache-2.0
-import type { PolicyModule, ReasonCode } from "../../types/policy.js";
+import type { PolicyEvaluationContext, PolicyModule, ReasonCode } from "../../types/policy.js";
 import type { DecisionInput, DecisionComputationResult } from "./types.js";
 import { deepMerge } from "../../utils/deepMerge.js";
 
 /**
  * runDecisionModules — the explicit decision phase of the OxDeAI policy engine.
  *
- * Evaluates a set of policy modules against (intent + state) and returns a
- * deterministic ALLOW / DENY decision with accumulated post-policy state.
+ * Evaluates a set of policy modules against (intent + state + trusted clock)
+ * and returns a deterministic ALLOW / DENY decision with accumulated
+ * post-policy state.
  *
  * Contract:
  *  - Pure: the input state is never mutated
  *  - Deterministic: identical inputs always produce identical outputs
- *  - No IO: no async, no external calls, no entropy
+ *  - No IO: no async, no external calls, no entropy. The trusted clock is
+ *    received via `input.evaluationTime`, never sampled here — that is what
+ *    keeps this phase replayable.
  *
  * Module evaluation semantics:
  *  - All modules are evaluated against the same pre-call working state.
  *    Modules do NOT see deltas from preceding modules during evaluation.
+ *  - All modules receive the same `PolicyEvaluationContext` instance, so the
+ *    trusted clock cannot drift between modules within one evaluation.
  *  - Results are processed in module-order.
  *  - ALLOW results whose stateDelta is non-null are deep-merged into the
  *    accumulating working state.
@@ -39,7 +44,11 @@ export function runDecisionModules(
   // Evaluate all modules against the current working state.
   // All modules see the same pre-delta state — cumulative inter-module
   // delta propagation during evaluation is intentionally not supported.
-  const results = modules.map((m) => m.evaluate(input.intent, working));
+  // The trusted evaluator context is built once and shared by every module,
+  // so all modules in one evaluation observe the same clock value.
+  const context: PolicyEvaluationContext = { evaluationTime: input.evaluationTime };
+
+  const results = modules.map((m) => m.evaluate(input.intent, working, context));
 
   // Process results in module-order: accumulate deltas on ALLOW,
   // collect reasons on DENY. In fail-fast mode, stop after the first DENY.
--- a/packages/core/src/policy/decision/types.ts
+++ b/packages/core/src/policy/decision/types.ts
@@ -7,12 +7,20 @@ import type { ReasonCode } from "../../types/policy.js";
  * Input to the decision phase.
  *
  * Contains only what is needed to make a deterministic authorization
- * decision: the proposed intent, the current policy state, and the
- * evaluation mode. No external IO or entropy is required.
+ * decision: the proposed intent, the current policy state, the trusted
+ * evaluation clock, and the evaluation mode. No external IO or entropy is
+ * required — `evaluationTime` is supplied by the caller rather than sampled,
+ * which is what keeps this phase pure and replayable.
  */
 export interface DecisionInput {
   intent: Intent;
   state:  State;
+  /**
+   * Trusted PEP clock (unix seconds), sampled once per evaluation by the
+   * caller and already validated by `assertProtocolSeconds`. Distinct from
+   * `intent.timestamp`, which is attacker-controlled.
+   */
+  evaluationTime: number;
   mode:   "fail-fast" | "collect-all";
 }
 
--- a/packages/core/src/policy/modules/ReplayModule.ts
+++ b/packages/core/src/policy/modules/ReplayModule.ts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import type { Intent } from "../../types/intent.js";
 import type { State } from "../../types/state.js";
-import type { PolicyResult } from "../../types/policy.js";
+import type { PolicyEvaluationContext, PolicyResult } from "../../types/policy.js";
 import { statelessModuleCodec } from "./_codec.js";
 
 function nonceKey(intent: Intent): string {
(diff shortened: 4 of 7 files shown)


Whatsonyourmind:
CI is green on all four checks. Correcting point 11 of my description, though, because I got the mechanism wrong and the answer to your fingerprint question turned out to be more interesting than "regenerate it".

**There is no fingerprint delta.** I rebuilt on this branch and the hash is byte-identical to the committed one:

```
committed : 36144da28664ee53c30b053fff5e79a0c91ea6f0e2195fd8676b970f0493c4ac
rebuilt   : 36144da28664ee53c30b053fff5e79a0c91ea6f0e2195fd8676b970f0493c4ac
```

Unlike #202 this isn't a Windows artifact — `packages/core/dist/index.d.ts` comes out pure LF here, so the value is canonical. Nothing to regenerate for #205.

I also said the check was "left for CI", which is wrong: there is no fingerprint step in `ci.yml` or `security-gate.yml`. It only runs via `pnpm -C packages/core api:fingerprint:check`.

**Why it didn't move is the part worth your attention.** `API_FINGERPRINT` is `sha256(packages/core/dist/index.d.ts)`, and that file is a 30-line barrel — `export type * from "./types/policy.js"` and friends, no inlined declarations. So the hash tracks the *set of export statements*, not the *shape* of what is exported.

This PR is a clean demonstration: it adds a new `@public` interface and widens a public method signature, both of which land in the emitted declarations —

```
dist/types/policy.d.ts:59  export interface PolicyEvaluationContext {
dist/types/policy.d.ts:65  evaluate(intent: Intent, state: State, context: PolicyEvaluationContext): PolicyResult;
```

— and the fingerprint is unchanged. As currently defined it won't catch adding or removing interface members, changing a method signature, or adding a named export inside a module that's already `export *`-ed. Removing a whole export line would still trip it.

**The artifact that does capture shape is `etc/core.api.md`, and it's stale repo-wide.** Running `pnpm -C packages/core api:report` regenerates it with 25 hunks / ~168 added lines, of which only about six are mine (`PolicyEvaluationContext`, and `evaluate` gaining its third parameter). The rest is accumulated drift from work merged since the report was last written at `c504838` — `AuthorizationV1` gaining `version`, `signature` becoming a union, `AuthorizationVerificationResult`, `authorizationSigningPayload`'s widened return type, and others.

I deliberately did **not** commit that regeneration. Folding ~160 lines of someone else's unreviewed API drift into #191 would break the isolation you asked for and would quietly bless changes that never got an API review. It reads to me like the same shape as #204 — a separate remediation that lands on its own and then everything downstream is clean. Happy to open that as its own PR if you want it, and equally happy to leave it to you; either way it seemed better to surface than to either ignore or silently absorb.

If you'd rather the guard actually track shape, hashing the full emitted `.d.ts` set — or wiring `api:check` (which fails on an out-of-date report rather than rewriting it) into CI — would close the blind spot. That's clearly outside #191, so I'll only raise it as an issue if you want it tracked.

AngeYobo:
Thanks, Luka. Your fingerprint analysis is correct.

Since `packages/core/dist/index.d.ts` is only a barrel containing `export type *` declarations, hashing that file does not protect the actual public API shape. Adding an interface member or changing a method signature inside an already re-exported module can leave the fingerprint unchanged, exactly as #205 demonstrates.

I agree with both decisions:

- Do not update `API_FINGERPRINT` in #205 because the canonical computed hash is unchanged.
- Do not regenerate the stale `etc/core.api.md` report in #205, because that would mix accumulated unrelated API drift into an otherwise isolated trusted-time change.

I will track the API guard weakness separately. That remediation should establish a reviewed API-report baseline and then make CI validate the actual declaration surface, either through API Extractor or through a deterministic hash of the complete emitted declaration set.

Before approval, I see two replay-specific points that should be corrected.

### 1. The replay-bypass direction is postdating, not backdating

The current commit message, module documentation, and one test say that a caller can backdate an intent to move `windowStart` before a recorded nonce and evict it.

Given the pruning rule:

```ts
keep entry when entry.ts >= windowStart
````

backdating moves `windowStart` earlier, which retains more entries. It does not evict the previously recorded nonce.

Postdating moves `windowStart` later and can prune an existing nonce, which is the replay-bypass direction. Backdating may extend retention or increase replay-state pressure, but it is not the eviction bypass described here.

The test:

```ts
"#191 replay window: a backdated intent cannot evict its own earlier nonce"
```

appears vacuous against the previous implementation because the recorded nonce would also be retained when `intent.timestamp` is heavily backdated.

Please replace it with a postdated replay test that would fail under the previous implementation. For example:

```ts
test("#191 replay window: a postdated intent cannot evict its previously recorded nonce", () => {
  const recorded = { nonce: "42", ts: EVAL_NOW - 60 };
  const postdated = intentAt(EVAL_NOW + 2 * WINDOW, 42n);

  const out = ReplayModule(
    postdated,
    stateWithNonces([recorded]),
    { evaluationTime: EVAL_NOW },
  );

  assert.equal(out.decision, "DENY");
  assert.deepEqual(out.reasons, ["REPLAY_NONCE"]);
});
```

Please also correct the attack description in the commit body and `ReplayModule` documentation.

### 2. The legacy-state convergence bound is not always exactly one retention window

A legacy entry may have been stamped from an `intent.timestamp` that was ahead of trusted `evaluationTime` by up to the configured future-skew tolerance.

Such an entry can survive for up to:

```text
replay retention window + permitted positive legacy timestamp skew
```

measured from the transition point.

Therefore this statement is too strong:

```text
after one replay retention window, every surviving entry carries a trusted-time stamp
```

The compatibility statement should instead say:

```text
Legacy entries self-heal as they expire under the trusted evaluation clock.
The transition completes within one replay retention window plus the maximum
positive legacy timestamp skew permitted by the freshness policy.
```

Alternatively, avoid an exact duration and state only that legacy entries are retained as-is and eventually age out under trusted-time pruning.

The architecture itself looks correct:

* `evaluationTime` is passed through `DecisionInput`;
* one `PolicyEvaluationContext` instance is shared across all modules;
* replay pruning and new nonce stamps use only `context.evaluationTime`;
* no ambient clock or authorization timestamp is introduced;
* the remaining modules receive the context but stay behaviorally unchanged.

Once the attack-direction test and compatibility wording are corrected, #205 should be ready for final review.

Whatsonyourmind:
Both corrections are in `0317969`. You were right on the direction, and checking it turned up a second instance of the same fault.

**1. Attack direction.** Retention is `entry.ts >= windowStart`, so backdating lowers `windowStart` and retains *more* — it cannot evict. Postdating is the eviction bypass. My test asserted `DENY` in the direction where the old implementation also returns `DENY`, so it proved nothing.

Rather than swap just that one, I replayed every window fixture against both the intent-clock and the trusted-clock rule to see which assertions actually separate them. **The first test was vacuous too**, for the mirror-image reason: it asserts a stale entry *is* pruned, but I had postdated the intent, and the old rule prunes it as well. That one needed backdating — the opposite fix to the one you asked for.

| test | asserts | old rule | new rule | |
|---|---|---|---|---|
| stale entry pruned (postdated, as written) | not retained | prunes | prunes | vacuous |
| stale entry pruned (**backdated**) | not retained | retains | prunes | separates |
| fresh entry retained (postdated) | retained | prunes | retains | separates |
| own nonce not evicted (backdated, as written) | DENY | DENY | DENY | vacuous |
| own nonce not evicted (**postdated**) | DENY | ALLOW | DENY | separates |

So each fixture now postdates where it asserts eviction is prevented, and backdates where it asserts pruning still happens. All three fail against the intent-clock implementation.

I also put the rule in the test file's header, since the direction is easy to get backwards — I did — and the failure mode is a green test rather than a red one.

**2. Convergence bound.** Adopted your wording. A legacy entry may carry an `intent.timestamp` running ahead of trusted time by up to the configured future-skew tolerance, so the overhang is real:

> Legacy entries self-heal as they expire under the trusted evaluation clock. The transition completes within one replay retention window plus the maximum positive legacy timestamp skew permitted by the freshness policy.

The module docstring and the commit body both say this now, and the docstring notes that the freshness gate is what makes the overhang finite rather than what removes it.

The commit message carries the corrected attack description as well — amended rather than appended, so the history doesn't preserve the wrong explanation.

**Re-validated after the change**, same gates as before:

| gate | result |
|---|---|
| core typecheck | clean |
| core tests | **296 pass / 0 fail** |
| conformance | **209 assertions** |
| vectors ts / auth / pep / delegation | **11 / 12 / 9 / 12** |
| `security:gate` | Blocking **0**, Decision **ALLOW** |
| `repro-policy-boundary-attacks` | byte-identical to `upstream/main` |
| lint | clean |
| `git diff --check` | clean |

Thanks for the API-guard call — establishing a reviewed report baseline first and then validating the actual declaration surface is the right order, and it keeps the drift out of this change. Happy to help on that once it has an issue, if useful.

AngeYobo:
Thanks, Luka. This is exactly the level of correction I was looking for.

The old-rule versus new-rule matrix is especially useful because it confirms the tests are genuinely discriminating rather than merely green.

The corrected fixture directions now make sense:

backdating for the stale-entry pruning case, where the old rule incorrectly retains the entry;
postdating for the fresh-entry retention case, where the old rule incorrectly prunes the entry;
postdating for the replayed nonce case, where the old rule can evict the recorded nonce and allow the replay.

The revised convergence wording is also precise and correctly attributes the finite migration bound to the freshness policy.

With the commit message, module documentation, and tests now aligned, I have no remaining blockers on #205.

The API compatibility guard remediation will remain separate under #206 Make the core public API compatibility guard shape-aware. That issue will establish a reviewed API report baseline and wire shape-aware API validation into CI. No fingerprint or API report update is required in #205.

Approved. Ready to merge once the final CI run on 0317969 is green.

PR #207, opened 2026-07-29
Fix/206 core api shape aware guard

### LINKED ISSUE #206: P1: Make the core public API compatibility guard shape-aware
## Summary

The current `@oxdeai/core` public API compatibility guard does not reliably detect changes to the shape of exported TypeScript declarations.

`packages/core/API_FINGERPRINT` is currently computed from:

```text
packages/core/dist/index.d.ts
````

That file is only a barrel containing re-export statements such as:

```ts
export type * from "./types/policy.js";
export type * from "./verification/types.js";
```

Because the declarations are not inlined into `dist/index.d.ts`, the fingerprint primarily tracks the set of barrel export statements, not the structure of the declarations behind those exports.

As a result, legitimate public API changes can leave the fingerprint unchanged.

## Confirmed blind spot

Issue #191 demonstrated the problem directly.

The change introduced a new public interface:

```ts
export interface PolicyEvaluationContext {
  evaluationTime: number;
}
```

and widened the public `PolicyModule.evaluate` signature:

```ts
evaluate(
  intent: Intent,
  state: State,
  context: PolicyEvaluationContext
): PolicyResult;
```

These changes appear in:

```text
packages/core/dist/types/policy.d.ts
```

but the current fingerprint remained byte-identical:

```text
committed:
36144da28664ee53c30b053fff5e79a0c91ea6f0e2195fd8676b970f0493c4ac

rebuilt:
36144da28664ee53c30b053fff5e79a0c91ea6f0e2195fd8676b970f0493c4ac
```

The current guard therefore does not detect at least the following classes of public API changes:

* adding or removing members from an exported interface;
* changing a public method signature;
* changing a public return type;
* adding a new declaration inside a module already re-exported with `export *`;
* widening or narrowing exported union types;
* changing optional properties to required properties, or vice versa.

It may still detect changes that modify the barrel itself, such as adding or removing an entire export line, but that is insufficient as an API compatibility control.

## API Extractor report drift

The repository also contains:

```text
etc/core.api.md
```

which is generated through API Extractor and does capture the public declaration shape.

However, the report is currently stale.

Running:

```bash
pnpm -C packages/core api:report
```

produces approximately:

```text
25 hunks
~168 added lines
```

Only a small subset is attributable to the latest change. The remainder is accumulated drift from previously merged public API changes, including examples such as:

* `AuthorizationV1` additions;
* signature type widening;
* `AuthorizationVerificationResult`;
* widened return types;
* other public declaration changes merged since the report baseline was last updated.

Regenerating this report inside an unrelated feature PR would silently approve accumulated API drift without a dedicated review.

## Objective

Make the public API compatibility guard shape-aware and ensure CI fails when the actual exported declaration surface changes without an explicit, reviewed baseline update.

## Required outcome

The repository must have a deterministic guard that detects public API shape changes across the complete emitted declaration surface, not only changes to the barrel file.

A public API change should require an explicit reviewed update before merge.

## Recommended approach

Use API Extractor as the primary compatibility guard.

Suggested model:

1. Establish a reviewed baseline for:

```text
etc/core.api.md
```

2. Add the API Extractor check to CI:

```bash
pnpm -C packages/core api:check
```

3. Ensure CI fails when the generated API report differs from the committed baseline.

4. Keep `api:report` as the explicit command for intentionally updating the baseline:

```bash
pnpm -C packages/core api:report
```

5. Require API report changes to be reviewed as part of any PR that intentionally modifies the public API.

## Optional fingerprint hardening

The existing fingerprint may be retained as a fast supplemental check, but it should either:

* hash all emitted public `.d.ts` files deterministically; or
* hash the canonical API Extractor report.

For example, a declaration-set fingerprint could include all relevant files under:

```text
packages/core/dist/**/*.d.ts
```

with:

* deterministic path ordering;
* normalized LF line endings;
* stable relative paths included in the hash input;
* generated/source-map declarations excluded where appropriate;
* no dependence on filesystem traversal order.

The fingerprint must not hash only `dist/index.d.ts`.

## Baseline review requirement

The current `etc/core.api.md` drift must be handled in a dedicated remediation PR.

That PR must:

* regenerate the report from current `main`;
* clearly separate accumulated baseline drift from new intended changes;
* document the major public API changes represented in the new baseline;
* receive explicit API-surface review;
* avoid unrelated runtime or protocol changes.

The baseline regeneration must not be hidden inside #191 or another unrelated feature PR.

## Implementation requirements

* Keep this work isolated from protocol, trusted-time, verification, and runtime behavior changes.
* Do not change `AuthorizationV1` wire format.
* Do not change canonicalization or signing payloads.
* Do not modify existing public APIs merely to reduce the report diff.
* Do not weaken or bypass API Extractor checks.
* Do not update the API baseline without reviewing the accumulated changes.
* Use Linux/CI as the authoritative generation environment.
* Normalize outputs so Windows and Linux builds do not produce different compatibility results.

## Acceptance criteria

* [x] `etc/core.api.md` is regenerated from current `main` and explicitly reviewed.
* [x] The accumulated public API drift is documented in the remediation PR.
* [x] `pnpm -C packages/core api:check` passes on the accepted baseline.
* [x] `api:check` runs in CI for changes affecting `packages/core`.
* [ ] CI fails when a public interface member is added without updating the baseline.
* [ ] CI fails when a public method signature changes without updating the baseline.
* [ ] CI fails when a public return type changes without updating the baseline.
* [x] CI fails when an exported union or property shape changes without updating the baseline.
* [x] A purely internal implementation change does not alter the API report.
* [ ] A test-only or documentation-only change does not alter the API report.
* [ ] Any retained `API_FINGERPRINT` is made shape-aware or clearly documented as supplemental only.
* [ ] Linux generation is deterministic.
* [ ] Repository typecheck passes.
* [ ] Core tests pass.
* [ ] Conformance validation passes.
* [ ] Security gate returns `ALLOW`.
* [ ] `git diff --check` is clean.

## Suggested regression test

Add a script or CI fixture that temporarily modifies an exported declaration, for example:

```ts
export interface PolicyEvaluationContext {
  evaluationTime: number;
  testOnlyPublicField?: string;
}
```

Then confirm that:

```bash
pnpm -C packages/core api:check
```

fails until the API report is explicitly regenerated.

A corresponding internal-only change should not fail the API check.

## Non-goals

* Changing public API semantics.
* Modifying protocol behavior.
* Updating trusted-time logic.
* Changing AuthorizationV1.
* Changing canonicalization or vectors.
* Introducing a release-versioning policy.
* Automatically approving API changes by regenerating reports in CI.
* Treating every source change in `packages/core` as a public API change.

## Priority

P1

The current guard can report success while public method signatures and exported type shapes change. This weakens compatibility review and allows public API drift to accumulate silently.

## References

* #191, which demonstrated that a public interface and method-signature change left the current fingerprint unchanged.
* #172 and #173, which intentionally changed verification result surfaces.
* `packages/core/API_FINGERPRINT`
* `packages/core/dist/index.d.ts`
* `etc/core.api.md`
* `packages/core/api-extractor.json`
* `scripts/api-fingerprint.sh`
* `scripts/check-api-fingerprint.sh`

AngeYobo:
## Summary

Replaces the previous public API compatibility guard with shape-aware API Extractor checks for `@oxdeai/core`.

The guard now detects changes to exported type structure, method signatures, return types, unions, and optionality, rather than checking only exported symbol names.

Closes #206.

## Changes

- Use API Extractor report comparison as the primary compatibility gate
- Commit and review the package-local API reports:
  - `packages/core/etc/core.api.md`
  - `packages/core/temp/core.api.md`
- Add CI checks for:
  - `api:check`
  - `api:guard:test`
  - `api:fingerprint:check`
- Add isolated regression fixtures proving the guard rejects:
  - exported interface member changes
  - public method signature changes
  - public return type changes
  - exported union changes
  - exported optionality changes
- Confirm internal helper and documentation-only changes remain compatible
- Verify repeated API report generation and normalized fingerprints are deterministic
- Preserve API Extractor CRLF report formatting through `.gitattributes`
- Retain the normalized SHA-256 API report fingerprint as a supplemental integrity check
- Update `tsx` to `4.23.1`, resolving `esbuild` to the patched `0.28.1` release

## Validation

- Core API report generation passed
- API compatibility check passed
- API guard regression suite passed
- Supplemental API fingerprint check passed
- Core typecheck passed
- Core tests passed: 296/296
- TypeScript conformance passed: 209 assertions
- Go portable harness passed: 28 assertions
- Python portable harness passed: 28 assertions
- Security authorization gate: ALLOW
- Security findings: 0
- Security warnings: 0


AngeYobo:
Make the @oxdeai/core public API compatibility guard shape-aware by using API Extractor report comparison as the primary CI gate. Add regression coverage for exported interface members, method signatures, return types, unions, and optionality, while allowing internal and documentation-only changes. Keep the normalized API report fingerprint as a supplemental integrity check, preserve report line-ending conventions, and update tsx to resolve the patched esbuild version. All core tests, conformance harnesses, API checks, and the security authorization gate pass. Closes #206.

Issue #194, opened 2026-07-21
P2: Execute trusted-time conformance vectors


AngeYobo:
<html>
<body>
<html><head></head><body><h2>Summary</h2><p>Wire the existing trusted-time conformance vector set into executable validation so that the trusted-time protocol is continuously verified across supported implementations.</p><p>The repository currently contains:</p><pre><code class="language-text">packages/conformance/vectors/trusted-time.json
</code></pre><p>with trusted-time cases such as:</p><pre><code class="language-text">tt-neg-*
tt-pos-*
</code></pre><p>However, the vector file is not currently loaded by the TypeScript conformance validator or the language harnesses. It therefore functions as draft test data rather than an enforced protocol contract.</p><p>This issue must:</p><ul><li><p>finalize the vector schema;</p></li><li><p>connect the vectors to executable harnesses;</p></li><li><p>validate their expected decisions and reason codes;</p></li><li><p>include them in normal CI and release validation;</p></li><li><p>establish parity across supported implementations.</p></li></ul><p>Refs #171.</p><hr><h2>Priority</h2><p><strong>P2</strong></p><p>The runtime freshness gate is implemented and tested in core, but conformance claims remain incomplete while the normative vector family is not executed.</p><p>This issue should land after the prerequisite trusted-time behavior represented by the vectors is implemented. It must not weaken vectors merely to match incomplete runtime behavior.</p><hr><h2>Problem statement</h2><p>A committed vector file is not conformance coverage unless:</p><ol><li><p>a validator loads it;</p></li><li><p>each case is executed against a defined implementation surface;</p></li><li><p>actual output is compared with expected output;</p></li><li><p>failures cause a non-zero process exit;</p></li><li><p>CI runs the validator;</p></li><li><p>supported language ports execute equivalent cases.</p></li></ol><p>The audit found no references to <code inline="">trusted-time.json</code> in the current conformance harness. Consequently, changes can regress trusted-time behavior while all existing conformance commands remain green.</p><hr><h2>Objective</h2><p>Make the trusted-time vectors an executable and release-blocking protocol test family.</p><p>After this issue:</p><pre><code class="language-bash">pnpm -C packages/conformance validate
[...]

Issue #193, opened 2026-07-21
P1: Mint AuthorizationV1 timestamps from evaluation_time


AngeYobo:
## Summary

Change AuthorizationV1 issuance so that `issued_at` and `expiry` are derived exclusively from trusted `evaluation_time`, not attacker-controlled `intent.timestamp`.

The current engine freshness gate validates that an intent timestamp remains within configured past/future bounds. However, after validation, authorization issuance still follows the legacy pattern:

```ts
issued_at = intent.timestamp;
expiry = intent.timestamp + authorizationTtlSeconds;
```

This leaves attacker-controlled time embedded in the signed authorization artifact. The freshness gate bounds the manipulation, but it does not satisfy the stronger trusted-time invariant that the issuer—not the agent—defines the authorization validity window.

The required issuance model is:

```text
issued_at = evaluation_time
expiry    = evaluation_time + effective_ttl
```

Refs #171.

---

## Priority

**P1**

The freshness gate blocks extreme future-dating attacks, but authorization timestamps remain derived from untrusted input. This affects the signed artifact’s validity interval and creates unnecessary coupling between intent freshness and authorization lifetime.

The independent P0 verifier-side future-`issued_at` defense should still be implemented. Issuance hardening and verification hardening are complementary controls.

---

## Current behavior

The current issuance path derives authorization timestamps from:

```text
intent.timestamp
```

The audit identified the equivalent logic around:

```text
packages/core/src/policy/PolicyEngine.ts
```

with current semantics approximately:

```ts
const issuedAt = intent.timestamp;
const expiry = intent.timestamp + authorizationTtlSeconds;
```

The repository also contains explicit tripwire tests preserving this behavior so it cannot change accidentally. Those tests must be deliberately updated as part of this reviewed change rather than bypassed or removed without replacement.

The current flow is therefore:

```text
agent supplies intent.timestamp
        ↓
freshness gate bounds timestamp
        ↓
engine copies timestamp into issued_at
        ↓
engine derives expiry from that timestamp
        ↓
both values are signed into AuthorizationV1
```

---

## Required invariant

For every successfully issued AuthorizationV1:

```text
authorization.issued_at === evaluation_time
```

and:

```text
authorization.expiry === evaluation_time + effective_ttl
```

The signed validity window must not be derived from:

* `intent.timestamp`;
* authorization timestamps supplied by a caller;
* policy state timestamps;
* module-local `Date.now()`;
* any other untrusted or ambient clock.

[...]

PR #208, opened 2026-07-29
fix(sdk): take verifier time from the trusted clock, not intent.timestamp

### LINKED ISSUE #195: P2: Stop using intent.timestamp as verifier time in the SDK client

Whatsonyourmind:
Closes #195.

Scoped as you asked: SDK-only verifier time. No AuthorizationV1 issuance (#193), no velocity-window behavior (#192), no trusted-time conformance execution (#194).

## 1. Exact SDK APIs changed

| Symbol | Change |
| --- | --- |
| `OxDeAIClient.verifyAuthorization(intent, authorization)` | Gains an optional third parameter `opts?: { verificationTime?: number }`. |
| `createGuard(...)` internal auth check | No signature change. |

## 2. Previous and new verifier-time semantics

Previously both SDK paths passed `intent.timestamp` as the engine's `now`.

Now `OxDeAIClient.verifyAuthorization` uses `opts.verificationTime ?? this.clock.now()`, and `createGuard` reuses the single `evaluationTime` sample it already takes for the freshness gate. This follows the existing SDK style rather than adding a new construction API — both entry points already accept a `ClockAdapter`, which is the trusted-time boundary the issue asks to make explicit (Option B over Option C).

**The defect is worse than "the intent can influence verifier time".** `expiry` is currently minted as `intent.timestamp + ttl`, so the old comparison was `intent.timestamp > intent.timestamp + ttl` — never true. SDK expiry enforcement could not fire at all, for any artifact, and an expired authorization verified as valid indefinitely. That is what the first regression test pins.

## 3. One acceptance criterion cannot be met in this issue, and I did not fake it

> Timestamp independence: hold `verificationTime = T` constant, evaluate with `intent.timestamp = A, B, C`, verification result remains identical.

Two problems with that as written:

- `timestamp` is in `INTENT_BINDING_FIELDS`, so it is covered by `intentHash`, which is checked *before* the expiry check. Varying only `intent.timestamp` therefore returns `AUTH_INTENT_MISMATCH` in every case — identically — **on the current unfixed code**. Written literally, that test passes before the fix and proves nothing.
- Issuing a matched authorization per timestamp avoids the hash mismatch, but then each artifact carries its own `intent.timestamp + ttl` deadline, so verification results still legitimately differ by intent timestamp. Strict independence is unreachable until `expiry` is minted from `evaluation_time` — which is #193, yours.

So I implemented the strongest property that is both true today and false before the change: **for a single trusted verification time past both deadlines, two authorizations issued at intent timestamps 190s apart are rejected, and rejected identically.** The verdict follows the trusted clock, not how recent each intent claims to be. I would rather flag this than ship a green check that does not test the thing.

## 4. Tests

`packages/sdk/src/sdk.test.ts`, two added:

- `client verification expires an authorization on the trusted clock, not intent.timestamp` — one authorization, byte-identical intent, only the trusted clock moves past expiry.
- `client verification result is independent of intent.timestamp` — the property above.

Both were **verified to fail against unfixed code** by reverting only `client.ts` / `guard.ts` and re-running: 6 pass / 2 fail, with `actual: { valid: true }` where an expired authorization was accepted. With the fix: 8 pass / 0 fail.

The assertions on `expiry` values are deliberate. I originally assumed expiry derived from evaluation time; it does not, and asserting it rather than assuming it is what surfaced section 3 instead of letting a vacuous test through.

## 5. The SDK test suite was not running at all

`packages/sdk/tsconfig.json` excludes `src/**/*.test.ts`, while the test script globbed `dist/**/*.test.js` — a path nothing could ever be emitted to. `pnpm -C packages/sdk test` exited 0 reporting `tests 0`, so `sdk.test.ts` and `guard.test.ts` had never executed.

Fixed the way `packages/guard` already does it: a `tsconfig.test.json` plus a `build:test` script. This is enabling work rather than drift — the "timestamp-independence tests pass" criterion is unreachable while the runner executes nothing. Switching it on surfaced **no** pre-existing failures: all 6 previously-dead tests pass unmodified.

One consequence I had to close: `guard` publishes `files: ["dist", "!dist/test", ...]` because its tests compile into `dist`. The SDK's tests live directly in `src/`, so they land in `dist/*.test.js` with no exclusion. Added `"!dist/**/*.test.*"`, verified with `npm pack --dry-run` — 21 files, no test artifacts, all real dist output still present.

## 6. Documentation and examples

- `packages/sdk/README.md`: new **Verification Time** section — the two clock domains and their trust, the explicit `verificationTime` form, and why deriving it from the intent means expiry can never be enforced.
- `packages/sdk/src/client.ts`: doc comment on the method, including that it inherits the limited engine-HMAC scope.
- `examples/agentgram-demo/src/agentgram.test.ts`: two `engine.verifyAuthorization(..., intent.timestamp)` call sites now pass the trusted `NOW` constant. Behaviour-identical there (`intent.timestamp === NOW`), 28/28 still pass — but it was the one place left demonstrating the pattern this issue removes, and criterion 4 is "examples no longer encourage incorrect verifier usage". **Happy to drop this file if you would rather keep the PR strictly inside `packages/sdk`.** I did not touch the `evaluatePure(intent, state, intent.timestamp)` calls in that file — issuance time is #193.

No doc or README anywhere showed the SDK client being called with `intent.timestamp`, so there was nothing else to correct.

## 7. Compatibility

- **Source compatible.** The new parameter is optional; every existing two-argument call compiles unchanged.
- **Semver:** minor by signature, but it is a **behavioural change** worth a release note. Authorizations that previously verified as valid forever will now correctly fail with `AUTH_EXPIRED` once the trusted clock passes their deadline. Callers who relied on the old result were relying on a check that could not fire.
- **No deprecation shim.** There is no overload documenting `intent.timestamp` as verifier time to keep alive, and adding a legacy path would preserve exactly the behaviour this issue exists to remove.

## 8. Conformance and core

Core verifier semantics unchanged — the diff touches no file under `packages/core`. Conformance: 209 assertions, unchanged.

## 9. Validation

| Gate | Result |
| --- | --- |
| `git diff --check` | clean |
| `pnpm typecheck` | pass |
| `packages/core` | 296 pass / 0 fail |
| `packages/sdk` | 8 pass / 0 fail (was 0 tests) |
| `packages/guard` | 145 pass / 0 fail |
| `packages/sift` | 196 pass / 0 fail |
| `packages/cli` | 26 pass / 0 fail |
| `packages/compat` | 10 pass / 0 fail |
| `tests` | 27 pass / 0 fail |
| `examples/agentgram-demo` | 28 pass / 0 fail |
| conformance validate | 209 assertions, pass |
| vectors ts / auth / pep / delegation | 11 / 12 / 9 / 12, all pass |
| `security:gate` | ALLOW, 0 blocking |
| `repro-policy-boundary-attacks` | exit 0, byte-identical across two runs |
| `git status --porcelain` | 0 |

One caveat, pre-existing and not from this branch: `pnpm test` at the root aborts on Windows because `@oxdeai/tests` runs `node --test $(find dist ... | sort)` and cmd.exe cannot parse POSIX command substitution. I confirmed it fails identically on clean `upstream/main` (exit 255, same message) and ran that package under bash instead to get the 27/27 above. Noting it only so the Windows-local failure is not mistaken for something this PR introduced; it needs no action here.

### DIFF
--- a/examples/agentgram-demo/src/agentgram.test.ts
+++ b/examples/agentgram-demo/src/agentgram.test.ts
@@ -337,11 +337,12 @@ test("decision can be verified independently", async () => {
 
   // Cast as Authorization: runtime object retains internal engine fields even
   // though the TypeScript type was narrowed to AuthorizationV1 at evaluatePure boundary.
+  // Verifier time is the trusted clock (NOW), never intent.timestamp.
   const verification = engine.verifyAuthorization(
     intent,
     result.authorization! as Authorization,
     result.nextState!,
-    intent.timestamp
+    NOW
   );
 
   assert.equal(verification.valid, true);
@@ -380,11 +381,12 @@ test("tampered intent fails verification", async () => {
   };
 
   // Cast as Authorization: same rationale as above — internal HMAC verification.
+  // Verifier time is the trusted clock (NOW), never intent.timestamp.
   const verification = engine.verifyAuthorization(
     tamperedIntent,
     result.authorization! as Authorization,
     result.nextState!,
-    intent.timestamp
+    NOW
   );
 
   assert.equal(verification.valid, false);
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -112,6 +112,29 @@ await guard(intent, async () => {
 });
 ```
 
+## Verification Time
+
+Authorization verification has two clock domains, and they are not interchangeable:
+
+| Value | Trust |
+| --- | --- |
+| `intent.timestamp` | Untrusted. Supplied by the agent. |
+| verifier time | Trusted. Supplied by the execution boundary. |
+
+The SDK takes verifier time from the trusted `clock` passed to `OxDeAIClient` and `createGuard`,
+never from the intent. A caller that already holds a trusted time from the execution boundary can
+pass it explicitly:
+
+```ts
+await client.verifyAuthorization(intent, authorization, {
+  verificationTime: trustedClock.now()
+});
+```
+
+Deriving verifier time from `intent.timestamp` compares an authorization against a deadline the
+same untrusted value produced, so expiry can never be enforced. `clock` defaults to
+`Date.now() / 1000`; supply a `ClockAdapter` when the host has an independently trusted time source.
+
 ## Main Exports
 
 - `buildIntent`, `buildState`
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -10,10 +10,12 @@
   "scripts": {
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "build": "rm -rf dist && tsc -p tsconfig.json",
-    "test": "pnpm build && node --test \"dist/**/*.test.js\""
+    "build:test": "rm -rf dist && tsc -p tsconfig.test.json",
+    "test": "pnpm build:test && node --test \"dist/**/*.test.js\""
   },
   "files": [
     "dist",
+    "!dist/**/*.test.*",
     "README.md",
     "CHANGELOG.md",
     "LICENSE"
--- a/packages/sdk/src/client.ts
+++ b/packages/sdk/src/client.ts
@@ -90,12 +90,32 @@ export class OxDeAIClient {
     return { output, state, auditEvents: emitted.events };
   }
 
-  async verifyAuthorization(intent: Intent, authorization: AuthorizationV1): Promise<{ valid: boolean; reason?: string }> {
+  /**
+   * Verifies an authorization against a trusted verification time.
+   *
+   * Verifier time is taken from the client's trusted clock boundary, or from an
+   * explicit `verificationTime` when the caller already holds one from the trusted
+   * execution boundary. It is never derived from `intent.timestamp`: that value is
+   * attacker-supplied and is hash-bound into the authorization at issuance, so using
+   * it as "now" pins the expiry comparison to issuance time and the authorization
+   * never expires.
+   *
+   * ⚠️ Inherits the LIMITED SCOPE of `PolicyEngine.verifyAuthorization` — it
+   * authenticates only the engine-HMAC field subset. Relying parties enforcing an
+   * authorization issued by an external party must use the strict standalone
+   * verifier with explicit `trustedKeySets` instead.
+   */
+  async verifyAuthorization(
+    intent: Intent,
+    authorization: AuthorizationV1,
+    opts?: { verificationTime?: number }
+  ): Promise<{ valid: boolean; reason?: string }> {
     const state = await this.stateAdapter.load();
+    const verificationTime = opts?.verificationTime ?? this.clock.now();
     // Cast as Authorization: internal engine.verifyAuthorization() uses legacy fields
     // (engine_signature, state_snapshot_hash) for HMAC binding verification.
     // Authorization objects returned by evaluatePure() retain these at runtime.
-    const result = this.engine.verifyAuthorization(intent, authorization as Authorization, state, intent.timestamp);
+    const result = this.engine.verifyAuthorization(intent, authorization as Authorization, state, verificationTime);
     return result.valid ? { valid: true } : { valid: false, reason: result.reason };
   }
 
--- a/packages/sdk/src/guard.ts
+++ b/packages/sdk/src/guard.ts
@@ -133,7 +133,10 @@ export function createGuard(opts: GuardOptions): GuardFn {
     if (opts.verifyAuthorization !== false) {
       // Cast as Authorization: runtime object retains internal engine fields even
       // though the TypeScript type was narrowed to AuthorizationV1 at the evaluatePure boundary.
-      const authCheck = opts.engine.verifyAuthorization(intent, out.authorization as Authorization, out.nextState, intent.timestamp);
+      // Verifier time is the single trusted `evaluationTime` sample taken above, never
+      // intent.timestamp — that value is attacker-supplied, so reading it as "now" would
+      // let the intent decide whether its own authorization has expired.
+      const authCheck = opts.engine.verifyAuthorization(intent, out.authorization as Authorization, out.nextState, evaluationTime);
       if (!authCheck.valid) {
         return {
           output: { decision: "DENY", reasons: [`AUTH_INVALID:${authCheck.reason ?? "unknown"}`] },
--- a/packages/sdk/src/sdk.test.ts
+++ b/packages/sdk/src/sdk.test.ts
@@ -7,6 +7,49 @@ import { PolicyEngine, RECOMMENDED_TRUSTED_TIME_PROFILE } from "@oxdeai/core";
 import { InMemoryAuditAdapter, InMemoryStateAdapter } from "./adapters.js";
 import { buildIntent, buildState } from "./builders.js";
 import { OxDeAIClient } from "./client.js";
+import { createGuard } from "./guard.js";
+
+/**
+ * Records the time argument the SDK hands to the engine, then delegates to the real
+ * implementation so behaviour is unchanged. This is what lets the tests below assert
+ * the verifier-time *argument* directly rather than inferring it from a verdict.
+ */
+class SpyEngine extends PolicyEngine {
+  readonly evaluateTimes: number[] = [];
+  readonly verifyTimes: Array<number | undefined> = [];
+
+  override evaluatePure(
+    ...args: Parameters<PolicyEngine["evaluatePure"]>
+  ): ReturnType<PolicyEngine["evaluatePure"]> {
+    this.evaluateTimes.push(args[2]);
+    return super.evaluatePure(...args);
+  }
+
+  override verifyAuthorization(
+    ...args: Parameters<PolicyEngine["verifyAuthorization"]>
+  ): ReturnType<PolicyEngine["verifyAuthorization"]> {
+    this.verifyTimes.push(args[3]);
+    return super.verifyAuthorization(...args);
+  }
+}
+
+function spyEngine() {
+  return new SpyEngine({
+    policy_version: "v1",
+    engine_secret: "test-secret-must-be-at-least-32-chars!!",
+    authorization_ttl_seconds: 60,
+    ...RECOMMENDED_TRUSTED_TIME_PROFILE
+  });
+}
+
+function testState() {
+  return buildState({
+    policy_version: "v1",
+    agent_id: "agent-1",
+    allow_action_types: ["PROVISION"],
+    allow_targets: ["us-east-1"]
+  });
+}
 
 test("builder helpers create engine-compatible intent/state", () => {
   const state = buildState({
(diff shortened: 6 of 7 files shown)


AngeYobo:
Thanks, this is a strong investigation. The finding that SDK expiry enforcement could never fire when verifier time was taken from intent.timestamp is important, and the scope separation from #193, #192, and #194 is correct.

The SDK test-runner repair is justified and should stay. Please also keep the Agentgram change, since it removes the remaining example of the incorrect verifier-time pattern.

Before merge, please make two small corrections:

1. Check the README Verification Time example. In the diff I reviewed, the TypeScript code fence appears not to be closed before the explanatory paragraph.

2. Strengthen the timestamp non-authority regression so it directly proves the SDK argument passed to the engine verifier. The current test shows that two artifacts are both rejected at a trusted time after both expiries, but it does not directly establish that intent.timestamp was not forwarded as verifier time.

Please add focused coverage, using a spy or equivalent controlled engine boundary, proving that:

- OxDeAIClient.verifyAuthorization passes opts.verificationTime when supplied;
- otherwise it passes this.clock.now();
- createGuard passes the single evaluationTime sample already used for evaluation;
- varying intent.timestamp does not change the verifier-time argument.

This remains fully inside #195 and avoids depending on #193 issuance semantics.

You can rename the current second test to describe what it actually proves, or replace it with the direct argument-capture test. Once those points are addressed and CI remains green, the PR is ready to merge.

Whatsonyourmind:
Thanks — and point 2 is a fair hit. The test proved the outcome, not the argument, and those are different claims. Fixed in `7ac0e24`.

**Argument capture.** Added a `SpyEngine extends PolicyEngine` that records the time argument and delegates to the real implementation, so behaviour is unchanged and the contract is asserted directly:

- `client forwards the trusted verification time to the engine verifier` — asserts the recorded arguments are exactly `[CLOCK_T]` with no options, then `[CLOCK_T, EXPLICIT_T]` once `verificationTime` is supplied, then that three further calls varying only `intent.timestamp` (−120s, +120s, +240s) append `[CLOCK_T, CLOCK_T, CLOCK_T]`. Those three fail the intent-hash check, which is deliberate and commented: the assertion is on what the SDK passed, not on the verdict it got back, so the hash mismatch cannot mask the property.
- `guard forwards its single evaluationTime sample to the engine verifier` — spies `evaluatePure` as well and asserts `evaluateTimes === verifyTimes === [CLOCK_T]` while the intent carries `CLOCK_T + 200`. That covers "the single sample already used for evaluation" rather than merely "some trusted value".

I kept the end-to-end test and renamed it to what it actually proves: `expired authorizations are rejected on the trusted clock regardless of intent timestamp`. It is weaker than the argument capture but it is the only one that exercises the real expiry path end to end, so it seemed worth keeping alongside rather than instead.

All four are verified to fail against the previous behaviour, not just to pass against the new one. Because the optional third parameter does not exist on the old signature, a straight revert would not compile, so I reverted `guard.ts` outright and reverted only the *body* of the client method (`opts?.verificationTime ?? this.clock.now()` → `intent.timestamp`), keeping the signature so the suite still built. Result: **6 pass / 4 fail**, with the six pre-existing tests unaffected. Restored: 10/10.

**On point 1 — I could not reproduce it, so I have not changed anything there.** The fence looks closed to me on the branch: in `packages/sdk/README.md` the block opens at line 128 with ` ```ts `, closes at line 132, and line 133 is blank before the paragraph. Rendering that section through GitHub's own markdown API returns `<pre>…</pre>` followed by a separate `<p>` for the explanatory paragraph, so the prose is not being swallowed into the code block.

I would rather ask than "fix" something that is already correct and make it worse. If it still renders wrong for you, tell me what you are seeing and I will restructure the section — my guess is the closing fence sits directly against the following paragraph in the diff view, which reads ambiguously with `+` prefixes even when the file is fine.

Keeping the test-runner repair and the Agentgram change as instructed.

**Validation after the amend:** `git diff --check` clean · typecheck pass · core 296/0 · sdk 10/0 · guard 145/0 · cli 26/0 · compat 10/0 · tests 27/0 · agentgram 28/0 · conformance 209 · security:gate ALLOW, 0 blocking · `git status --porcelain` 0. Single commit, DCO signed.

AngeYobo:
Thanks, this addresses both review points.

You were correct about the README fence. The source and rendered structure are valid, so no change is required there.

The SpyEngine coverage is the right correction. It now proves directly that:

- the client uses clock.now() by default;
- explicit verificationTime takes precedence;
- varying intent.timestamp does not alter the verifier-time argument;
- createGuard reuses the same evaluationTime sample for evaluation and verification.

One final adjustment before merge: please remove the exact expiry assertions and commentary that preserve the current #193 behavior:

assert early.expiry === intent timestamp + 60
assert late.expiry === intent timestamp + 60

Those assertions encode the issuance defect that #193 is about to change and would make this SDK suite fail as soon as trusted-time issuance lands.

Keep the end-to-end test, but make it issuance-model-neutral:

- read both actual expiry values from the returned authorizations;
- choose TRUSTED = Math.max(earlyExpiry, lateExpiry) + 1;
- verify both artifacts at that trusted time;
- assert both return AUTH_EXPIRED;
- do not assert whether expiry was derived from intent.timestamp or evaluation_time.

Please also rename ISSUED_AT to EVALUATION_TIME in that test where it represents the trusted clock sample rather than necessarily the artifact issued_at value.

After that amendment and the existing validation remaining green, the PR is ready to merge.

Whatsonyourmind:
Good catch, and a better one than mine. I added those assertions to keep the test discriminating, and in doing so pinned the suite to the exact issuance behaviour #193 is about to change — the test would have gone red on your own next PR, for a reason that has nothing to do with what it is testing. Done in `12e7fd6`:

- Both exact `expiry` assertions and the `#193` commentary are gone.
- `TRUSTED = Math.max(earlyExpiry, lateExpiry) + 1`, read from the returned authorizations. The test now asserts only that both are `AUTH_EXPIRED` at a single trusted time past both deadlines, and makes no claim about how either deadline was derived.
- `ISSUED_AT` → `EVALUATION_TIME`. I applied the rename in the first test as well, since the constant means the same thing there (the trusted clock sample, not the artifact's `issued_at`) — say the word if you wanted it confined to the end-to-end test.

I re-checked that this did not quietly defang it: reverting the client method body to `intent.timestamp` and `guard.ts` to its previous state still fails all four, 6 pass / 4 fail, and restoring gives 10/10. Removing an assertion is exactly the kind of edit that can turn a real test into a vacuous one, so it seemed worth confirming rather than assuming.

Validation unchanged and green: typecheck · core 296/0 · sdk 10/0 · guard 145/0 · cli 26/0 · compat 10/0 · agentgram 28/0 · conformance 209 · security:gate ALLOW · `git diff --check` clean · `git status --porcelain` 0. Still one commit, DCO signed.

AngeYobo:
Thanks Luka, this resolves the final review point. The broader EVALUATION_TIME rename is correct, so please keep it. The end-to-end test is now issuance-model-neutral, and the explicit regression check against the previous implementation confirms the suite remains discriminating. With all checks green and DCO signed, this is approved for merge.

PR #231, opened 2026-08-08
fix(core): reclaim expired concurrency leases and remove released ones

### LINKED ISSUE #227: P2: Expired active_auths leases are never reclaimed
## Summary

`concurrency.active_auths[agent_id].expires_at` is validated and included in the state hash, but current policy evaluation does not use it to reclaim an expired concurrency slot.

An `EXECUTE` whose matching `RELEASE` never arrives can therefore retain its slot indefinitely until authoritative state is manually rewritten.

## Current behavior

- `expires_at` is present and codec-validated.
- The concurrency module uses `active_auths` for allocation and release bookkeeping.
- No evaluation path removes an entry solely because its trusted expiration time has passed.
- A missing `RELEASE` can leave the agent at its concurrency limit indefinitely.

## Security and operational impact

This creates a persistent availability failure:

- a crashed executor;
- a lost release message;
- an ambiguous execution outcome;
- or an interrupted workflow

may permanently consume a concurrency slot.

This is a lease-lifecycle issue, not an approval-workflow issue.

## Relationship to other work

- Related to #197 because automatic or operator-initiated reclamation must mutate authoritative state through the approved versioned provider boundary.
- Independent from #218, which concerns approval-state and workflow semantics.
- Must use trusted `evaluationTime`, not an intent-controlled timestamp.

## Decision required

Determine whether expired leases are reclaimed:

1. deterministically during the next policy evaluation;
2. by a separately authenticated operator or maintenance operation;
3. by provider-managed expiration;
4. or through a combination with precisely defined ownership.

Any reclamation path must preserve exact-version conflict detection and must not permit execution after a CAS conflict.

## Acceptance criteria

- Expired leases cannot hold concurrency capacity indefinitely.
- Expiration uses trusted evaluation time.
- Reclamation ownership is explicitly defined.
- State mutation uses the authoritative provider and version domain.
- Concurrent reclaim, release, and execute operations fail closed on version conflict.
- Conformance tests cover abandoned leases, boundary timestamps, and racing mutations.

Whatsonyourmind:
Closes #227. Implements ruling **(b)** from https://github.com/oxdeai/oxdeai/issues/227#issuecomment-5213888429, with the clean-RELEASE non-removal included as you directed.

## Shape

`ConcurrencyModule` keeps the decision. `PolicyEngine` materializes the resulting `concurrency.active_auths[agent]` map as a replacement assignment after module evaluation, inside the same evaluated transition and the same exact-version CAS, so `active` and `active_auths` move atomically. `deepMerge` is untouched, no tombstones, no generic deletion semantics.

The two call sites share one exported function, `computeConcurrencyLeases(intent, state, evaluationTime)`, so the decision has exactly one definition. The engine calls it with `state` — the same pre-delta input the module evaluated against — and the same trusted `evaluationTime`, so engine and module cannot drift apart.

## Against your requirements

| Requirement | Where |
|---|---|
| reclamation uses only the trusted `evaluationTime` | `computeConcurrencyLeases`; the module reads `context.evaluationTime`, the engine passes the same value |
| remove only leases the concurrency evaluation determined reclaimable | engine assigns the map that function returned; it does not re-derive one |
| no mutation outside the evaluated transition/CAS | assignment sits between the decision phase and `state_snapshot_hash`, on `working` |
| no tombstones | none |
| no change to generic `deepMerge` semantics | `deepMerge` is no longer referenced by `PolicyEngine` at all, and is otherwise unmodified |
| clean RELEASE removes its entry | `T-11` |
| `active` consistent with the resulting map | decremented by exactly the number removed, floored at zero |
| invariant: no released/reclaimed lease resident after a successful transition | `INV-ConcurrencyLeases` |

Rulings 2, 3 and 5 from your earlier comment are implemented as issued: expiry is `evaluationTime >= expires_at`, malformed `expires_at` fails closed with `STATE_INVALID` with no maximum-plausible-TTL ceiling, and a late RELEASE stays a deterministic `CONCURRENCY_RELEASE_INVALID`.

## Two judgement calls worth your eye

**1. `active` is decremented, not recomputed from the map size.** Ruling 4 says decrement by exactly the number removed, so that is what it does — but it is worth naming why the alternative is wrong. A deployment that maintains `active` without populating `active_auths` would have its counter silently reset to the map size, which would *release* capacity that was never lease-tracked. Decrementing leaves such a deployment exactly as it was.

**2. Only the acting agent's leases are read.** Ruling 3 says a malformed `expires_at` fails closed, and the scope it applies at is mine to choose. Reading only `active_auths[intent.agent_id]` means one corrupt entry denies that agent rather than every agent sharing the state. If you would rather a malformed leaf anywhere fail the whole evaluation closed, say so and I will widen it — it is a two-line change, but it converts one bad entry into a full outage, so I did not assume it.

## Coverage

`T-11`–`T-18` in the TOCTOU suite, covering all six cases from ruling 6 plus the two you added: clean RELEASE preserving unrelated leases, reclamation during EXECUTE, the exact `expires_at` boundary in both directions, saturation recovery after abandoned leases, late RELEASE after reclaim, malformed `expires_at` (NaN / Infinity / non-integer / negative, on both EXECUTE and RELEASE), eligibility, and no double decrement. Plus `INV-ConcurrencyLeases` in the invariants suite.

I reverted `ConcurrencyModule.ts` and `PolicyEngine.ts` to `cef2d72` and re-ran with the tests in place: **all 8 new TOCTOU tests fail, all 10 pre-existing ones pass, and both invariants fail.** Every new test depends on the change rather than passing incidentally.

`T-6` carried an "Observation (not asserted)" recording the old `deepMerge` limitation. That observation is now false, so it has been rewritten to point at `T-11`, which asserts the removal directly.

**On `T-18`**, so the name does not overclaim: core's `StateStore` is a plain `get`/`set` and has no CAS to exercise. The test asserts the property that makes the Profile C CAS sufficient — reclamation is a pure function of `(intent, state, evaluationTime)`, so two evaluators racing from one version compute the *same* transition rather than two stacking ones, and the loser re-evaluating against the winner's state is a `DENY`. The CAS itself is the provider's obligation, which is what the §2.2 paragraph now states.

## Spec

`state-provider-requirements` §2.2 gains one paragraph on lease reclamation, parallel in structure to the existing velocity and tool-call window paragraphs, plus a line pointing at §4.3 for the privileged recovery path so it is not read as the normal one.

## Verification

- `packages/core` **354/354** (346 before, +8)
- `tests` **32/32** (30 before, +2)
- `packages/guard` **145/145**
- monorepo `pnpm typecheck` — 0 errors
- portable vectors — canonicalization 11, authorization 12, PEP 9, delegation 12, all unchanged
- `check-api-fingerprint.sh` OK; `api-extractor` produces no diff, since `policy/modules/` is not re-exported from the package entry point

Consistent with your note that no `AuthorizationV1`, canonicalization, or portable artifact-vector change was expected.

### DIFF
--- a/docs/integrations/policy-state-modeling.md
+++ b/docs/integrations/policy-state-modeling.md
@@ -216,20 +216,25 @@ replay and concurrency only — so release paths are not gated on the same polic
 surface as execution, and a release cannot be starved by, say, an exhausted
 budget.
 
-One property to model around: **`active_auths[...].expires_at` is stored and
-validated, but nothing evicts on it.** It is carried through the state codec and
-contributes to the state hash, and no module reads it to reclaim a slot. A
-long-running action whose `RELEASE` never arrives therefore retains its
-concurrency slot under current runtime behavior.
+`active_auths[...].expires_at` is evicted on. A lease is expired when
+`evaluationTime >= expires_at`, and expired leases are reclaimed during policy
+evaluation rather than by a background sweeper, so a long-running action whose
+`RELEASE` never arrives stops holding its concurrency slot once the lease
+expires. Reclamation uses trusted `evaluationTime` and the same authoritative
+provider and version domain as guard transitions, and the counter and the lease
+map move together in the transition that triggered it, committed through the
+same exact-version CAS ([#227](https://github.com/oxdeai/oxdeai/issues/227)). The
+lease problem remains distinct from the approval/workflow gap tracked by #218.
 
-Expired-lease reclamation is tracked separately by
-[#227](https://github.com/oxdeai/oxdeai/issues/227); its final owner and algorithm
-remain unresolved. Any future automatic, operator-driven, or provider-managed
-reclamation uses trusted `evaluationTime` and the same authoritative provider
-and version domain as guard transitions. It must commit with an exact expected
-version through CAS or equivalent conflict rejection. This guidance does not
-implement reclamation, and the lease problem remains distinct from the
-approval/workflow gap tracked by #218.
+One property to model around: **`active` is an authoritative input, not a value
+derived from `active_auths`.** State must satisfy `active[agent] >=` the number
+of resident entries in `active_auths[agent]`. A counter *above* the tracked
+lease count is intentionally valid — it can represent capacity you account for
+but never lease-track, and reclamation decrements rather than recomputes so that
+capacity survives. A counter *below* it fails closed with `STATE_INVALID`.
+Because a missing `active[agent]` reads as `0`, populating `active_auths`
+without setting `active` denies every call for that agent: set `active`
+explicitly when adopting lease tracking.
 
 ### State versioning
 
--- a/docs/spec/state-provider-requirements.md
+++ b/docs/spec/state-provider-requirements.md
@@ -88,6 +88,25 @@ Concurrent evaluators can still race to consume the same remaining tool quota,
 so providers MUST serialize or reject conflicting updates. An evaluation clock
 behind a persisted call timestamp fails closed with `STATE_INVALID`.
 
+The same requirements apply to concurrency leases. A lease in
+`concurrency.active_auths` is expired when `evaluation_time >= expires_at`, and
+expired leases are reclaimed deterministically during policy evaluation rather
+than by a background sweeper. Reclamation removes the expired entries and
+decrements `concurrency.active` by exactly the number of entries removed, so the
+counter and the lease map MUST be written in the same transition as the
+operation that triggered the reclamation — a provider that commits one without
+the other leaves capacity permanently mis-accounted. Concurrent evaluators can
+read the same leases and each reclaim them, so providers MUST serialize or
+reject all but one through atomic compare-and-set; because reclamation is a pure
+function of the state read and the trusted `evaluation_time`, the rejected
+evaluator recomputes the same transition against the winner's state rather than
+applying a second decrement. A malformed `expires_at` leaf — missing,
+non-finite, non-integer, or negative — fails closed with `STATE_INVALID`.
+
+A separately authenticated privileged provider mutation remains available as an
+operational recovery path (§4.3); it is not the normal path, and deployments
+MUST NOT rely on it to release capacity in ordinary operation.
+
 ### 2.3 Stale replica reads
 
 For Profile C deployments, `getState()` MUST read from the authoritative replica of the state store, not from a potentially stale secondary replica. Reading from a lagging replica can produce a hash that matches an old authorization but not the current authoritative state.
--- a/docs/spec/state-provider-requirements.md
+++ b/docs/spec/state-provider-requirements.md
@@ -103,6 +103,19 @@ evaluator recomputes the same transition against the winner's state rather than
 applying a second decrement. A malformed `expires_at` leaf — missing,
 non-finite, non-integer, or negative — fails closed with `STATE_INVALID`.
 
+The counter is an authoritative input, not a value inferred from the lease map.
+Before any reclamation is applied, state MUST satisfy `concurrency.active[agent]
+>= the number of resident entries in concurrency.active_auths[agent]` for the
+acting agent. A counter above the tracked lease count is intentionally valid — a
+deployment may account capacity in `active` that it never lease-tracked, and
+reclamation decrements rather than recomputes so that capacity keeps being
+counted — while a counter below it offers capacity the resident leases have
+already consumed, and fails closed with `STATE_INVALID`. Because a missing
+`concurrency.active[agent]` entry reads as `0`, a provider that populates
+`active_auths` without setting `active` is under-counted by this rule and its
+agents fail closed; deployments adopting lease tracking MUST set `active`
+explicitly as a migration step rather than relying on it being inferred.
+
 A separately authenticated privileged provider mutation remains available as an
 operational recovery path (§4.3); it is not the normal path, and deployments
 MUST NOT rely on it to release capacity in ordinary operation.
--- a/packages/core/src/policy/PolicyEngine.ts
+++ b/packages/core/src/policy/PolicyEngine.ts
@@ -10,7 +10,6 @@ import { canonicalJson, intentHash, sha256HexFromJson } from "../crypto/hashes.j
 import { engineSignHmac } from "../crypto/sign.js";
 import { engineVerifyHmac } from "../crypto/verify.js";
 import { SIGNING_DOMAINS, signEd25519, signHmacDomain } from "../crypto/signatures.js";
-import { deepMerge } from "../utils/deepMerge.js";
 import { runDecisionModules } from "./decision/index.js";
 
 import { HashChainedLog } from "../audit/HashChainedLog.js";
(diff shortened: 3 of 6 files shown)


AngeYobo:
Hi Luka, 
The overall shape looks right and (b) is implemented the way I intended. Keeping one `computeConcurrencyLeases()` definition and using the engine only to materialize the replacement map is the correct boundary.

I agree with your first judgement call that `active` should **not** simply be recomputed from `active_auths.size`: deployments may conservatively account capacity in `active` that is not lease-tracked, and recomputing would silently release that capacity.

There is one asymmetry I want closed before merge, though.

We need to fail closed when the scalar counter is **lower than the number of tracked resident leases**.

For example:

```text
max_concurrent = 1
active = 0
active_auths = { live-A }
```

Today an EXECUTE would see `activeAfterReclaim = 0`, ALLOW a new lease, and end with two live tracked leases while `max_concurrent = 1`.

So please add an authoritative-state consistency check before applying reclamation:

```text
active >= number of tracked resident leases
```

must hold for the acting agent.

Do **not** require equality - `active > tracked leases` remains valid/conservative for deployments that account untracked capacity. But:

```text
active < tracked leases
→ DENY / STATE_INVALID
```

This also means we should not need `Math.max(0, active - removed)` once the invariant is established; `removed <= resident leases <= active`.

Please add regression coverage for the under-count case (EXECUTE at minimum, RELEASE too if it shares the same validation path) and extend the invariant suite accordingly.

On your second judgement call: keep validation scoped to the acting agent. I do not want one malformed lease belonging to an unrelated agent to turn into a global state-wide denial.

Everything else in the current shape looks aligned with #227.

Whatsonyourmind:
Done in `040ab5d`.

`computeConcurrencyLeases` now evaluates the precondition before any reclamation:

```text
active >= number of tracked resident leases   (acting agent)
```

`active < tracked` → `STATE_INVALID`. Equality is not required — `active > tracked` stays valid and the untracked capacity survives the transition rather than being normalized away. Validation stays scoped to the acting agent, as you ruled.

The check sits in `computeConcurrencyLeases` rather than in the module so engine and module keep one definition of the decision. `PolicyEngine` reaches the lease materialization only after a non-DENY decision, so a rejected state cannot produce a partial write.

You were right that the floor then becomes unnecessary — `removed <= resident <= active` — so `Math.max(0, active - removed)` is gone. Keeping it would have masked exactly the state the precondition rejects.

RELEASE does share the validation path, so it is covered too. Worth naming the resulting precedence: when the counter is under-counted **and** the RELEASE names a genuinely resident lease, the result is `STATE_INVALID`, not `CONCURRENCY_RELEASE_INVALID` — the state precondition runs first. That reads correct to me (the lease is fine; the state is not), but it is a visible ordering choice, so flagging rather than assuming.

**One migration consequence to rule on.** `active` reads as `state.concurrency.active?.[agent] ?? 0`, so a deployment that populates `active_auths` but never sets `active` now fails closed on every call for that agent. That is literally `active < tracked`, so it is inside your rule as written, and I have not carved an exemption for it. Flagging it because it is the one shape that previously worked and now denies — say the word if you want it treated differently.

Coverage:

- **T-19** — EXECUTE under-count fails closed; still rejected with spare `max_concurrent` (the defect is the state, not the limit); over-count still ALLOWs with the untracked slot preserved
- **T-20** — RELEASE under-count fails closed, with the precedence above
- **INV-ConcurrencyLeases** — over-count valid / under-count invalid / consistent-at-limit still `CONCURRENCY_LIMIT_EXCEEDED`

Local: core 356/356, tests 33/33, guard 145/145, conformance 17/17, typecheck clean, API fingerprint OK, portable vectors unchanged.

I neutralized the new guard and re-ran before trusting any of it: T-19, T-20 and the new invariant fail without it, and all 354 pre-existing core tests pass either way — so the tests exercise the new behaviour and the change does not move anything that was already covered.

AngeYobo:
Luka, this looks correct.
Keep the precedence as implemented:
under-counted state + resident RELEASE
 -> STATE_INVALID
The RELEASE target is valid; the authoritative concurrency state is not.
On the migration consequence, do not add an exemption for `active_auths` populated with missing `active`. With the current model, missing `active` means `0`, so that state is under-counted and should fail closed.
I would rather require an explicit state migration than infer `active` from `active_auths`, because `active > tracked` is intentionally valid and may represent untracked capacity.
Please document that migration requirement in the state-provider/concurrency guidance so operators know that Tier 1 state must satisfy:
active >= tracked resident leases
With that documented, I consider the blocking review resolved.

Whatsonyourmind:
Documented in `b7a2215`. Precedence kept as implemented, and no exemption added — missing `active` stays under-counted and fails closed.

`state-provider-requirements` §2.2 now carries the requirement:

> state MUST satisfy `concurrency.active[agent] >= the number of resident entries in concurrency.active_auths[agent]` for the acting agent

with the asymmetry stated explicitly — above the tracked count is intentionally valid because it can represent untracked capacity and reclamation decrements rather than recomputes; below it fails closed — and the missing-`active` case written up as an explicit migration step rather than something the engine infers.

One thing I corrected while I was there, since this PR falsifies it. `docs/integrations/policy-state-modeling.md` still described the pre-#227 runtime:

> **`active_auths[...].expires_at` is stored and validated, but nothing evicts on it.** … A long-running action whose `RELEASE` never arrives therefore retains its concurrency slot under current runtime behavior.

…with reclamation "tracked separately" and its "final owner and algorithm remain unresolved". That section now describes eviction as implemented and carries the counter requirement in operator-facing terms.

**Flagging rather than editing:** `docs/architecture/decisions/tier1-evaluator-input-provenance.md` carries the same pre-#227 framing — "Its final owner and algorithm remain unresolved… This ADR does not implement or resolve reclamation." That is #197's document and you are mid-revision on it after the rulings you posted there, so I have left it alone rather than editing underneath you. Happy to send it as a separate one-paragraph PR if you would rather not carry it, but it seemed better folded into your own pass.

Verified after the docs change: core 356/356, working tree limited to the two doc files.

AngeYobo:
Thanks. This resolves my blocking concern.

The invariant, precedence, and migration behavior are now documented
consistently, and leaving the #197 ADR untouched was the right call since
that document is already under active revision.

I’ll fold the stale reclamation wording in
tier1-evaluator-input-provenance.md into the #197 pass rather than
splitting it into another PR.

From my side, #231 is ready to merge once required CI is green.

Whatsonyourmind:
Heads-up before you merge: CI on `b7a2215` came back **3 green / 1 red** — `adapter-validation` failed with `Envelope verification failed: invalid` on the `langgraph` example (`openai-tools` passed first). `build-test`, `security` and `pep-redis-replay` are green.

I'm reporting it rather than calling it a flake, because I could not prove either way. What I did rule out:

- **The commit is docs-only.** `b7a2215` touches two `.md` files; `adapter-validation` passed on `040ab5d`, which carries the identical code. Nothing under `scripts/`, `.github/workflows/`, or any build input references either doc path.
- **The new precondition cannot deny that example.** `examples/langgraph/src/policy.ts` seeds `concurrency.active = {}` and `active_auths = {}`, so the check evaluates `0 < 0` and passes, and the pair stays consistent as the run's state evolves (one EXECUTE → `active 1`, one resident lease).
- The failing assertion is `verifyEnvelope(..., mode: "strict")` returning a non-`ok` status, i.e. the envelope itself, not a policy decision.

What I could not do: I don't have admin on the repo so I can't re-run the job, and the validator's own `pnpm --filter @oxdeai/core build` step fails on my Windows box (my platform, not the repo — a direct `pnpm --filter @oxdeai/core build` succeeds), so I can't reproduce that adapter locally.

A re-run on the same SHA would settle it. If it reproduces, it's a real red and I'll chase it — say the word and I'll dig rather than guess.

AngeYobo:
Thanks for isolating it this carefully. Agreed, let’s not classify it as a flake without evidence.
I’ll re-run adapter-validation on the same b7a2215 SHA.
If it passes unchanged, we can treat the previous failure as non-deterministic CI noise. If it reproduces, please dig into the langgraph strict envelope verification path before we merge.
No merge while the required check is red.

AngeYobo:
Re-run on the same b7a2215 SHA is green. adapter-validation passed unchanged, so this looks like non-deterministic CI noise rather than a regression from your docs commit.

All required checks are green now. Thanks for flagging it instead of assuming a flake.

Issue #233, opened 2026-08-11
P1: Add Tier 1 secure guard entry point and trusted execution context


AngeYobo:
## Summary:

Implement the first production boundary from #197 without changing legacy `PolicyEngine` behavior.

The secure guard path must construct trusted execution context server-side after authentication and must not deserialize authoritative provenance fields directly from proposer-controlled request JSON.

## Initial scope:

 * dedicated Tier 1 secure guard entry point
* `TrustedExecutionContext` or equivalent internal structure
* trusted principal, agent binding input, adapter/tool route identity, security-relevant route classification, depth, tenancy posture, and evaluation time
* generic trusted-vs-proposer conflict handling
* deterministic fail-closed behavior on conflicting claims
* legacy direct `PolicyEngine` path remains available but carries no Tier 1 provenance guarantee

## Out of scope:

* authoritative StateProvider implementation
* PolicyProvider implementation
* argument provenance
* credential provenance
* default-deny tool semantics
* changes to `AuthorizationV1` or canonicalization

 Closes no umbrella issue. Implements the first bounded step of #197.

Whatsonyourmind:
Taking this one. Proposed shape below, nothing written to a branch yet — same process as #227.

I re-read the #197 rulings and the ADR before designing rather than working from my own summary of the review, which changed my first instinct: the ADR puts the context "outside core `PolicyEvaluationContext`" and lists moving it into core under Alternatives Considered, so this stays a guard/PEP concern and core is untouched.

## Where it lives

`packages/guard`, additive. `OxDeAIGuard` keeps its current behavior and signature — it stays the legacy/low-level path with no Tier 1 provenance guarantee, per "Secure and Legacy Profiles". The secure path is a second entry point sharing one enforcement body, so there is one PEP with two ways of establishing provenance, not two PEPs that can drift.

The concrete gap it closes is visible today: `defaultNormalizeAction` does `requireContextString(action.context, "agent_id")`, i.e. authoritative identity is read out of proposer-supplied JSON.

## Entry point

```ts
// Illustrative.
const secure = createSecureGuard(config);
await secure(trustedContext, proposal, execute, opts);
```

Trusted context is a separate positional argument, never a key inside `ProposedAction`. "Cannot be supplied as request JSON" then holds structurally rather than by convention — there is no field on the proposal type to put it in.

## Construction

[...]

Issue #215, opened 2026-07-31
P2: ActionType is a closed union; operator-defined action types cannot be expressed


Whatsonyourmind:
Found while documenting policy/state modeling patterns for #166.

## What an operator wants

To gate an action the protocol does not enumerate - for example `DELETE_RECORD`, `SEND_MESSAGE`, `ROTATE_KEY`.

## What exists today

`ActionType` in `packages/core/src/types/intent.ts` is a closed union:

```ts
export type ActionType = "PAYMENT" | "PURCHASE" | "PROVISION" | "ONCHAIN_TX";
```

`intent.action_type` and `state.allowlists.action_types` are both typed by it.

## Consequence

An operator whose domain is not payments, purchasing, provisioning or on-chain transactions has to overload one of the four names, which makes the allowlist a proxy for something it does not describe and weakens the audit record's meaning.

## Scope

No protocol semantics change is proposed here, and none is made in the #166 documentation PR.

Issue #214, opened 2026-07-31
P2: No default-deny tool allowlist - allowlists has no tool axis


Whatsonyourmind:
Found while documenting policy/state modeling patterns for #166.

## What an operator wants

"This agent may call only these tools; refuse anything else."

## What exists today

`state.allowlists` has `action_types`, `assets` and `targets`. `AllowlistModule` reads those three and nothing else - there is no tool axis.

Tool constraints live in `state.tool_limits`, which expresses *rate*: `max_calls` per agent and `max_calls_by_tool` per agent per tool, over `window_seconds`.

## Why default-deny is not expressible

Per-tool *refusal* works: `max_calls_by_tool[agent][tool] = 0` makes the check `toolCount + 1 > 0` always true, so the tool is denied. That yields a **denylist**, which requires enumerating every tool to be blocked.

The default-deny direction is unavailable. `isRateLimitedToolCall` returns `false` - and therefore ALLOW - when neither `max_calls_by_tool[agent][tool]` nor `max_calls[agent]` is configured, so a tool absent from configuration is permitted rather than refused.

## Scope

No protocol semantics change is proposed here, and none is made in the #166 documentation PR. Recording the gap rather than approximating it, per that issue's acceptance criteria.

PR #234, opened 2026-08-12
feat(guard): add Tier 1 secure guard entry point and trusted execution context

Whatsonyourmind:
Implements #233, the first bounded step of #197. Shape and first diff were reviewed on the issue before this branch existed.

## What this adds

A second entry point in `@oxdeai/guard` that establishes trusted evaluator inputs before evaluation. `OxDeAIGuard` keeps its signature and behavior and remains the legacy/low-level path with no Tier 1 provenance guarantee.

`createSecureGuard` **delegates to `OxDeAIGuard`** rather than reimplementing enforcement, so state read, evaluation, authorization verification, replay consumption, intent/state hash binding, CAS and execution ordering are the same code on both paths and cannot drift.

Trusted context is a **separate positional argument**, never a field on `ProposedAction` — so "trusted context is not accepted from request JSON" holds structurally rather than by convention. `createTrustedExecutionContext` is the only constructor and sets a non-enumerable runtime brand; a type-only brand is erased at runtime and would let `JSON.parse(body).ctx` through.

The concrete gap this closes is present today: `defaultNormalizeAction` reads `agent_id` out of proposer-supplied `action.context`.

## Rulings from the issue, as implemented

- **Missing claims** — `absent → fill from the derived premise`, `present-and-equal → use the premise`, `conflict → fail closed`. Per-field outcomes surface on `GuardDecisionRecord.provenance`, so a derived fill and a correct proposer assertion are distinguishable in audit.
- **`unverified`** — a fourth outcome for a claim with no non-declared premise. Without it, a tool claim on a route with no tool identity would record identically to a verified match. `P-14` pins it.
- **Reason codes** — guard-layer only. `OxDeAIProvenanceConflictError extends OxDeAIAuthorizationError`. Core `ReasonCode`, the conformance registries and the API fingerprint are untouched.
- **Route classification** — carried and conflict-checked, deliberately not wired into policy-module evaluation. `intent.action_type` stays separate and #215 independent.
- **Depth** — required, validated, never defaulted. A test asserts `depth: 0` is still accepted, so "required" cannot later be misread as "non-zero".
- **`evaluationTime`** — not on the context; captured at the evaluation boundary. The ADR's illustrative shape is corrected in this PR, narrowly and with no other ADR change.
- **Brand** — documented as misuse resistance, explicitly not authentication. A privileged in-process caller reaching the builder can still manufacture a branded context.
- **Exact equality** — no case folding or normalization; a case-only difference is a conflict, with a test pinning it.
- **Extensibility** — the seam is named and structured in terms of provenance modes. `declared` and `derived` are implemented; `read` and `attested` attach later as sibling resolvers without restructuring, under `declared never overrides derived, read, or attested`.

## Attack coverage

`L` (privileged `agent_id` substitution) is closed on the secure path. `M` (tool rename) is closed **only** for routes supplying authoritative tool identity; where the route has none, the claim records `unverified` and no closure is claimed. Default-deny tool semantics remain #214.

## Verification

- **Control first:** `packages/guard` on `main` (`bb5ae92`) = **145 tests, 145 pass**.
- **After:** **164 tests, 164 pass** — 19 added, none of the 145 broken.
- **Neutralized and re-ran**, since tests that pass against the old code prove nothing: disabling conflict detection fails **8** tests including the L and M cases; disabling the derived fill fails **P-5**. Probes removed and green re-confirmed.
- `packages/core` **356/356**; `scripts/check-api-fingerprint.sh` passes unchanged.

## Out of scope

Authoritative StateProvider/CAS, PolicyProvider, credential and argument provenance, `tool_call` hardening, default-deny tool semantics, `AuthorizationV1` and canonicalization. The stale lease-reclamation wording in the ADR is left for the separate #197 documentation pass.

## One naming question

I kept the exported `reconcileWithTrustedContext` name as reviewed, since the enforcement context is the only non-declared source today. If you would rather generalise it now that `read` and `attested` are on the roadmap, say the word and I will rename it in place.

### DIFF
--- a/docs/architecture/decisions/tier1-evaluator-input-provenance.md
+++ b/docs/architecture/decisions/tier1-evaluator-input-provenance.md
@@ -154,9 +154,16 @@ interface TrustedExecutionContext {
   adapterId: string;
   tool?: string;
   depth: number;
-  evaluationTime: number;
 }
 ```
+
+The context deliberately does not carry `evaluationTime`. Trusted evaluation
+time is captured by the secure guard at the evaluation boundary, immediately
+before evaluation, following the existing trusted-time model, rather than being
+carried in a reusable authenticated context. Authentication establishes who and
+where; `evaluationTime` establishes when a specific decision is being made, so a
+context constructed once must not become a reusable time capsule for later
+decisions.
 
 The context is created server-side and must not be accepted directly from
 request JSON. The exact interface and construction rules belong to a follow-up
--- a/packages/guard/package.json
+++ b/packages/guard/package.json
@@ -32,7 +32,7 @@
     "typecheck": "tsc -p tsconfig.json --noEmit",
     "build": "rm -rf dist && tsc -p tsconfig.json",
     "build:test": "rm -rf dist && tsc -p tsconfig.test.json",
-    "test": "pnpm build:test && node --test \"dist/test/guard.test.js\" \"dist/test/guard.property.test.js\" \"dist/test/guard.delegation.test.js\" \"dist/test/guard.delegation.matrix.test.js\" \"dist/test/guard.delegation.property.test.js\" \"dist/test/guard.toctou.test.js\" \"dist/test/guard.boundary.test.js\" \"dist/test/guard.replay-store.test.js\" \"dist/test/guard.replay-store.redis.test.js\" \"dist/test/guard.authorization.test.js\" \"dist/test/guard.state-binding.test.js\" \"dist/test/guard.cas.test.js\" \"dist/test/guard.pep-conformance.test.js\" \"dist/test/guard.intent-binding.test.js\""
+    "test": "pnpm build:test && node --test \"dist/test/guard.test.js\" \"dist/test/guard.property.test.js\" \"dist/test/guard.delegation.test.js\" \"dist/test/guard.delegation.matrix.test.js\" \"dist/test/guard.delegation.property.test.js\" \"dist/test/guard.toctou.test.js\" \"dist/test/guard.boundary.test.js\" \"dist/test/guard.replay-store.test.js\" \"dist/test/guard.replay-store.redis.test.js\" \"dist/test/guard.authorization.test.js\" \"dist/test/guard.state-binding.test.js\" \"dist/test/guard.cas.test.js\" \"dist/test/guard.pep-conformance.test.js\" \"dist/test/guard.intent-binding.test.js\" \"dist/test/guard.provenance.test.js\""
   },
   "files": [
   "dist",
--- a/packages/guard/src/errors.ts
+++ b/packages/guard/src/errors.ts
@@ -63,6 +63,33 @@ export class OxDeAIConflictError extends OxDeAIAuthorizationError {
 }
 
 /**
+ * Thrown when a proposer claim contradicts a trusted execution-context premise
+ * on the Tier 1 secure guard path.
+ *
+ * Extends OxDeAIAuthorizationError so existing catch blocks remain valid. This
+ * is a guard-layer error by design: the Tier 1 provenance boundary is enforced
+ * at the PEP, and representing it here keeps core `ReasonCode`, the public
+ * conformance registries, and the API fingerprint unchanged.
+ *
+ * Raised before evaluation, authorization issuance, state mutation, or
+ * execution. `fields` lists every conflicting field in declaration order, and
+ * `provenance` carries the full per-field outcome for the audit record.
+ */
+export class OxDeAIProvenanceConflictError extends OxDeAIAuthorizationError {
+  readonly fields: readonly string[];
+  readonly provenance: Readonly<Record<string, string>>;
+
+  constructor(fields: readonly string[], provenance: Readonly<Record<string, string>>) {
+    super(
+      `Proposer claim conflicts with trusted execution context on [${fields.join(", ")}]. Execution blocked.`
+    );
+    this.name = "OxDeAIProvenanceConflictError";
+    this.fields = Object.freeze([...fields]);
+    this.provenance = Object.freeze({ ...provenance });
+  }
+}
+
+/**
  * Thrown when DelegationV1 verification fails at the guard boundary.
  * Extends OxDeAIAuthorizationError so existing catch blocks remain valid.
  * The `violations` field carries structured delegation-specific failure codes.
--- a/packages/guard/src/guard.ts
+++ b/packages/guard/src/guard.ts
@@ -12,6 +12,7 @@ import {
   OxDeAIDelegationError,
   OxDeAIGuardConfigurationError,
   OxDeAINormalizationError,
+  OxDeAIProvenanceConflictError,
 } from "./errors.js";
 
 // ── validation ────────────────────────────────────────────────────────────────
@@ -147,6 +148,11 @@ export function OxDeAIGuard(config: OxDeAIGuardConfig) {
       intent = normalize(action);
     } catch (err) {
       if (err instanceof OxDeAINormalizationError) throw err;
+      // The secure path performs trusted-vs-proposer reconciliation inside the
+      // normalizer, so a provenance conflict surfaces here. It is an
+      // authorization boundary failure, not a normalization failure, and must
+      // reach the caller with its own type and conflicting-field list intact.
+      if (err instanceof OxDeAIProvenanceConflictError) throw err;
       // Custom mapActionToIntent threw something unexpected — fail closed.
       throw new OxDeAINormalizationError(
         `mapActionToIntent threw an unexpected error: ${err instanceof Error ? err.message : String(err)}`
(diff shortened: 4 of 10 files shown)


AngeYobo:
Luka, I reviewed the diff and did another pass against the Tier 1 rulings. The overall implementation is strong, and the test neutralization checks hold up. I have three points before I consider #233 ready.

First, there is one real mismatch with the missing-claim ruling.

On the default secure path, agent_id cannot actually be absent and filled from trusted context today. defaultNormalizeAction requires action.context.agent_id and throws before reconciliation runs.

So this currently works:

trusted tool absent from proposal
-> normalize
-> reconciliation fills trusted tool

but this does not:

trusted agentId available

* proposer omits agent_id
    -> defaultNormalizeAction throws
    -> reconciliation never runs

That means the ruling:

claim absent
-> use trusted premise

is not actually true end-to-end for the main identity field.

Please fix this rather than documenting it as an exception.

The secure path should allow an absent proposer agent_id to be supplied from the trusted context for normalization purposes, while still running reconciliation against the original proposer claims after mapActionToIntent.

The important properties must remain:

* an absent proposer agent_id audits as absent
* a matching proposer agent_id audits as matched
* a conflicting proposer or custom-normalizer agent_id fails closed
* the custom normalizer still cannot overwrite the trusted identity

Please add an end-to-end test for:

trustedContext.agentId = A
proposal contains no agent_id
-> secure path evaluates as A
-> provenance[“agent_id”] = “absent”

Second, provenance conflicts currently do not reach onDecision.

That is understandable structurally because the conflict occurs inside normalization before the shared guard emits a decision record, and the structured exception does preserve the conflict fields and provenance.

But this differs from the ruling that the distinction should be preserved in the audit record. For a deployment using onDecision as its audit sink, attempted identity/tool substitutions are currently invisible there.

I do not want you to silently invent a synthetic policy decision just to solve this.

Please either:

1. provide a guard-level audit emission for provenance rejection with clearly separate semantics from a PolicyEngine decision, if there is already a clean place for it, or
2. keep conflicts exception-only in this PR and document the contract explicitly:

ALLOW / evaluated paths
-> provenance available on GuardDecisionRecord

pre-evaluation provenance conflict
-> provenance available on OxDeAIProvenanceConflictError, not on onDecision

If option 2 is cleaner, I am fine tracking unified guard-boundary audit emission separately. The important thing is that we do not imply onDecision contains these rejected attempts when it does not.

Third, the deployment-level tenancy ruling still needs to be represented.

We explicitly decided that tenantId may be absent only for an explicitly configured single-tenant deployment.

For multi-tenant deployments:

missing trusted tenantId
-> fail closed

proposer tenant != trusted tenant
-> fail closed

Right now tenantId is simply optional on TrustedExecutionContext, so the secure path cannot distinguish a legitimate single-tenant deployment from a multi-tenant integration that forgot to supply tenant identity.

Please add an explicit secure-path tenancy posture, or an equivalent deployment-level configuration, and tests for:

single-tenant + no tenantId
-> valid

multi-tenant + no trusted tenantId
-> fail before evaluation

multi-tenant + trusted tenantId + matching proposer tenant
-> valid

multi-tenant + trusted tenantId + conflicting proposer tenant
-> provenance conflict

Everything else in the diff looks aligned: the runtime brand semantics, evaluationTime removal, exact comparison, unverified, route classification handling, replay-store lifetime, shared enforcement body, and conditional M claim are all in the right shape.

Please keep the changes uncommitted for one more review after these points are addressed.

Whatsonyourmind:
All three addressed. Changes are **uncommitted** on the branch as you asked — `4b3ee23` is still the pushed head, and the working tree holds four modified files.

## 1. Absent `agent_id` — you were right, and it was a real gap

Confirmed against source before changing anything: `defaultNormalizeAction` calls `requireContextString(action.context, "agent_id")`, which throws before reconciliation runs. So "absent claim → derived premise" genuinely was not reachable end-to-end for identity.

Worth naming why I missed it: **P-5 exercised the fill rule on `tool`** — the one field where an absent claim survives normalization. The rule was tested where it happened to work rather than where it mattered.

Fixed rather than documented. The secure path now supplies the derived `agentId` to the normalizer **only when the proposer omitted it**, and reconciles against the **original** proposer claims, so the fill is not laundered into a claim the proposer made. All four properties hold:

- absent proposer `agent_id` → evaluates as the trusted identity, audits `absent` (`P-19`)
- matching proposer `agent_id` → audits `matched` (`P-20`)
- conflicting proposer `agent_id` → fails closed (`P-12`, unchanged)
- a custom normalizer that substitutes an identity into a synthesized field → fails closed (`P-21`)

That last one needed care. A synthesized field is still compared, because the guard handed the normalizer the premise and expects it echoed back exactly; a deviation is a substitution attempt, not a fill. Only the *recorded outcome* changes — a synthesized field that matches records `absent`, since the record describes the proposer's claim, not the normalizer's echo of my own value. Dropping the comparison instead would have silently overwritten a substituting normalizer rather than rejecting it.

## 2. Conflicts and `onDecision` — option 2, documented

I took option 2. There is no clean existing place for a guard-level audit emission: `onDecision` is typed to `GuardDecisionRecord` with `decision: "ALLOW" | "DENY"`, so emitting a conflict there means either a synthetic decision — which you ruled out, and which would misreport a boundary rejection as a policy outcome — or widening a public type, which is scope this PR should not take.

The contract is now stated explicitly on `createSecureGuard`:

```text
ALLOW / evaluated paths   → provenance on GuardDecisionRecord (onDecision)
pre-evaluation conflict   → provenance on OxDeAIProvenanceConflictError, NOT on onDecision
```

with the consequence spelled out — a deployment auditing only through `onDecision` will **not** see attempted identity or tool substitutions and must also record the exception. `P-25` pins the current behaviour so the documented contract and the code cannot drift apart. Happy for unified guard-boundary audit emission to be tracked separately.

## 3. Tenancy posture

`createSecureGuard(config, { tenancy })` is now required and **not defaulted**. Defaulting to single-tenant would recreate exactly the ambiguity the ruling removes — a multi-tenant integration that has not wired tenant identity would be indistinguishable from a legitimate single-tenant one. There are no existing callers of this entry point, so requiring it costs nothing.

- single-tenant + no `tenantId` → valid (`P-22`)
- multi-tenant + no trusted `tenantId` → fails closed before evaluation, execution and state mutation (`P-23`)
- multi-tenant + trusted `tenantId` + matching proposer tenant → valid, audits `matched` (`P-24`)
- multi-tenant + trusted `tenantId` + conflicting proposer tenant → provenance conflict (`P-24`)

## Verification

- `packages/guard` **164 → 171**, all passing; none of the earlier tests broken.
- **Neutralized each new behaviour and re-ran**, since three new tests that pass against the new code prove nothing on their own:
  - disable the `agent_id` synthesis → **P-19 and P-21 fail**
  - record synthesized fields as `matched` instead of `absent` → **P-19 fails**
  - disable the multi-tenant fail-closed → **P-23 fails**
  - probes removed, `grep` confirms none left, suite re-confirmed **171/171**
- `packages/core` **356/356**; `check-api-fingerprint.sh` passes unchanged; typecheck clean.

Ready for the next review pass whenever you are.

AngeYobo:
Thanks. This addresses all three review points cleanly.

The agent_id synthesis behavior is the right shape: supply the derived identity only to satisfy normalization, but preserve the original proposer claim set for provenance and still reject a custom normalizer that substitutes a different identity. P-19/P-21 cover the important distinction.

Option 2 for pre-evaluation conflicts is also correct. Do not manufacture a synthetic PolicyEngine DENY just to make onDecision fire. The documented contract is acceptable for this PR:

evaluated path -> provenance on GuardDecisionRecord

pre-evaluation provenance rejection -> provenance on OxDeAIProvenanceConflictError

We should track unified guard-boundary audit emission separately so deployments do not have to compose decision and exception streams forever, but that is not a blocker here.

The explicit tenancy posture is also correct. Requiring the deployment to choose single-tenant or multi-tenant is preferable to any default that could silently weaken the boundary.

171/171 guard, 356/356 core, typecheck clean, fingerprint unchanged, plus the neutralization probes are sufficient for this review pass.

Please commit the working tree, push it to #234, and let the required CI run. From my side, the implementation review is approved pending green CI.

AngeYobo:
Thanks. I’m ratifying the public shape as implemented:

`createSecureGuard(config, { tenancy })`

with public `SecureGuardOptions`.

The tenancy posture is part of the secure-profile contract, not an implementation detail, so making it explicit and mandatory is the right API choice. Do not add a default.

The re-checks confirm the review findings are resolved, including the end-to-end absent `agent_id` path and the pre-state-read multi-tenant fail-closed behavior.

The exception-only provenance contract for pre-evaluation conflicts is accepted for #234 as documented. I’ve opened #235 ,  the unified guard-boundary audit follow-up separately, rather than widening `onDecision` or manufacturing synthetic `PolicyEngine` decisions in this PR.

From my side, the implementation is approved.

Issue #262, opened 2026-09-02
P1: Fix broken pre-freeze security demos and verifier fixtures


AngeYobo:
## Summary

Fix shipped OxDeAI examples whose documented behavior no longer matches current 2.0 runtime requirements.

This is release-blocking because these examples are part of the public security evidence for the 2.0 boundary. Several currently fail for configuration or fixture reasons unrelated to the behavior they are supposed to demonstrate.

## Scope

Only:

- `examples/delegation-demo/**`
- `examples/non-bypassable-openclaw/**`
- `examples/rust-verifier/**`

Do not modify production package APIs unless a separate real runtime defect is discovered.

### Current failure

Both documented entry points fail immediately with:

```text
OxDeAIGuardConfigurationError: trustedKeySets are required
````

The three `OxDeAIGuard(...)` constructions provide `expectedAudience` but omit required `trustedKeySets`. 

### Required fix

* configure valid `trustedKeySets` for all guard instances
* mirror the current trusted-key pattern used by working sibling demos
* preserve existing delegation semantics
* do not weaken strict verification to make the demo pass
* verify both documented entry points actually run

### Current failure

`agent.mjs` hand-builds an incomplete/unsigned AuthorizationV1 instead of using the properly signed authorization fixture used by the working non-bypassable demo.

As a result, ALLOW / hash mismatch / replay scenarios fail for unrelated reasons such as:

* `AUTH_ISSUER_MISMATCH`
* `AUTH_MISSING_FIELD`

rather than demonstrating their intended boundary behavior. 

### Required fix

* reuse the canonical signed fixture/helper from the working non-bypassable demo where appropriate
* do not fabricate a reduced AuthorizationV1 shape
* preserve explicit ALLOW, intent/state mismatch, replay, and bypass scenarios
* ensure each scenario fails or succeeds for the documented reason
* keep bypass semantics scoped to the reviewed enforcement boundary
* if the example duplicates the existing non-bypassable demo without adding distinct OpenClaw evidence, document whether it should remain or be retired in a separate follow-up rather than silently deleting it here

### Current failure

The bundled authorization fixture has expired. The README says the bundled case should ALLOW, but the current verifier correctly returns:

```text
DENY (AUTH_EXPIRED)
```

The fixture expiry is historical and has become a time bomb. 

### Required fix

Prefer a deterministic fixture strategy that does not silently rot with wall-clock time.

Acceptable approaches include:

[...]

PR #276, opened 2026-09-04
ci(rust-verifier): run cargo test in CI so the fixture cannot rot silently

May-1600:
Closes the last open acceptance criterion on #262: *"fixture cannot silently expire again without CI detecting it."*

#264 landed the deterministic `VERIFY_NOW` path and the four fixture tests, but nothing runs `cargo test` in CI, so that criterion was still unenforced. This wires it up.

## Changes

- **`.github/workflows/ci.yml`** — new `rust-verifier` job, modeled on the existing Go 1.25 setup: stable toolchain via `dtolnay/rust-toolchain`, cargo cache via `Swatinem/rust-cache` scoped to the example, `cargo test --locked` with `working-directory: examples/rust-verifier`.
- **`package.json`** — root `test:rust` script, listed next to the other harness suites.
- **`examples/rust-verifier/README.md`** — points at the root entry point and the CI job.

`--locked` is deliberate: the example ships a `Cargo.lock`, and CI should fail on drift rather than silently resolve new versions.

## Proof the signal is real

A CI job is only worth adding if it actually goes red when the expiry contract rots. Verified by temporarily breaking the contract in `src/verify_authorization.rs` and observing the failures. Both mutations were reverted; no verifier source is changed by this PR.

**Mutation 1 — verifier ignores the trusted verification time and reads the wall clock.** This is the original defect from the issue: the bundled fixture's fixed expiry is in the past, so the canonical ALLOW case rots.

```
test allow_at_last_valid_second ... FAILED
  assertion `left == right` failed: expected ok, got
  [Violation { code: "AUTH_EXPIRED", message: "authorization has expired" }]
test deny_on_tampered_field ... FAILED
  left: "AUTH_EXPIRED"  right: "AUTH_SIGNATURE_INVALID"
test result: FAILED. 2 passed; 2 failed
```

**Mutation 2 — expiry boundary relaxed from exclusive to inclusive (`<=` to `<`).**

```
test deny_at_expiry_boundary ... FAILED
  left: "ok"  right: "invalid"
test result: FAILED. 3 passed; 1 failed
```

Both exit `101`, so the job fails the build.

## Local run

`cargo 1.98.0`, `rustc 1.98.0`:

```
Running unittests src/main.rs
test tests::absent_uses_real_now ... ok
test tests::valid_i64_overrides_real_now ... ok
test tests::malformed_is_explicit_error ... ok
test result: ok. 3 passed; 0 failed

Running tests/verify_authorization.rs
test allow_at_last_valid_second ... ok
test deny_at_expiry_boundary ... ok
test deny_when_expired ... ok
test deny_on_tampered_field ... ok
test result: ok. 4 passed; 0 failed
```

## Constraints

- No production package source changed. Scope is the example's README, the workflow, and the root `package.json`.
- Fixture expiry not moved into the future, as the issue requires.
- No verification check weakened or bypassed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,27 @@ jobs:
         env:
           OXDEAI_ENGINE_SECRET: ci-adapter-validation-secret-testing!!
 
+  rust-verifier:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Rust stable
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: examples/rust-verifier
+
+      - name: Rust verifier tests
+        # Pins the bundled fixture's ALLOW/DENY cases at deterministic protocol
+        # time (issue #262). Fails if the fixture or the expiry contract rots.
+        run: cargo test --locked
+        working-directory: examples/rust-verifier
+
   pep-redis-replay:
     runs-on: ubuntu-latest
     services:
--- a/examples/rust-verifier/README.md
+++ b/examples/rust-verifier/README.md
@@ -62,6 +62,11 @@ Expected outcomes in general:
 cargo test
 ```
 
+From the repo root, `pnpm test:rust` runs the same suite with `--locked`. CI runs
+it on every push and pull request (`rust-verifier` job in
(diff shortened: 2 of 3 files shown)


AngeYobo:
Thanks @May-1600 . I will review and merge the PR #276 asap. please do not close and reopen it

AngeYobo:
@May-1600 Main advanced by one docs-only commit after #277. Please update #276 with current main, without changing the PR scope. Once the required checks rerun green, I’ll merge it.

